PR #83, opened 2020-09-28
#5: Removing Old Backend Goyave Files

DPigeon:
# Related Issue
#5 

# Proposed changes
- Removed the old backend files.

# Additional Info
- While looking at the backend, I saw that he old backend Goyave files were not removed. Since we've moved from Goyave to Java, it must be removed.

# Reviewer(s)
 - @razine-bensari 
- @Alessandro100


razine-bensari (approved):
Nice Catch!

Issue #71, opened 2020-09-22
Setup Linter


Alessandro100:
As a developer I would like to have a linter in my project so that the code stays consistently formatted

**Note**
Use husky for pre-commit

**Acceptance Criteria**
- A linter is setup in the frontend 
- A linter is setup in the backend
- The rules of the linter will abide by our coding guidelines*
- The linters will prevent code from being commit if the code doesn't follow it

DPigeon:
The husky depency has been added on the frontend. A pre-commit will run the lint & unit tests before commiting on the frontend.
![setup-linter-frontend](https://user-images.githubusercontent.com/37888675/94953234-303fa080-04b5-11eb-83ff-d66d16f2fadc.png)

DPigeon:
Unfortunately, we cannot use husky with the frontend and backend at the same time. Husky will work with the frontend. However, to be able to have a pre-commit linter on the backend as well, the pre-commit hooks have to be overwritten. 

I'm currently looking for a solution for this. We could achieve this by creating some pre-commit shell script but everyone would have to replace their pre-commit hooks script with the one created.

DPigeon:
Update: To bypass the husky limitation, I have written a small script. Everything will be documented in a PR.

PR #86, opened 2020-10-04
#71: Setup Linter

DPigeon:
# Related Issue
#71 

# Proposed changes
- Removed Husky dependency on the frontend
- Created a new shell script to be put in your hidden directory **.git/hooks/** (PROPOSED SOLUTION)

# Additional Info
I had installed the husky dependency on the frontend to have a pre-commit linter strategy. However, this does not scale well with the backend because we're using Java. So I removed it and created a shell script. Since Git hooks can only be modified locally when using them without a dependency or a package, follow the steps below so that everyone has a working linter on both client and backend.

**Steps**
1. cd into the root directory of the project
2. Run the command
```
ln -s -f ../../hooks/pre-commit .git/hooks/pre-commit && chmod +x ./hooks/pre-commit
```
3. The new linter hook should now be configured to run with the project!

# Checklist (if applicable)
- [x] Tests

# Reviewer(s)
 - @Alessandro100 
 - @marwan011 

PS: I added everyone as a reviewer because everyone needs to follow the steps mentioned in *Additional Info*. If you ever find a better solution than this @Alessandro100 or @marwan011 let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,5 +7,5 @@ A web platform that encourage good investing practices.
 1. cd into the root directory of the project
 2. Run the command
 ```
-ln -s -f ../../hooks/pre-commit .git/hooks/pre-commit
+ln -s -f ../../hooks/pre-commit .git/hooks/pre-commit && chmod +x ./hooks/pre-commit
 ```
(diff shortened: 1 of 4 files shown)


Alessandro100 (approved):
Look good, I think we got the cleanest solution here

Issue #98, opened 2020-10-15
Setup frontend architecture


Alessandro100:
As a developer I would like to setup the frontend architecture so that it sets a good base for future work

**Acceptance Criteria**
- NGRX store is setup
- The design patterns are implemented
- The models are implemented
- The services are implemented

PR #112, opened 2020-10-20
Feature/setup frontend architecture #98

Alessandro100:
# Related Issue
- [link to the issue on zenhub](https://github.com/DPigeon/Money-Tree/issues/98)

# Proposed changes
- Implemented NgRx Store
- Implemented Services
- Implemented Interfaces (models)
- Implemented design patterns

# Additional Info
Design Patterns Implements
- Decorators (Component, Injectable): This is how Angular works
- Facade: When working with the NgRx Store we will use a Facade
- Dependency Injection: We will inject dependencies in the presentation layers to improve single responsibilities

# Checklist (if applicable)
- [X] Tests
- [X] Documentation

# Screenshots (if applicable)

Angular Architecture Diagram (1)

# Reviewer(s)
 - @razine-bensari 
 - @EspressoCode 
 - @arthur-del 
 - @marwan011

### DIFF
--- a/client/src/app/interfaces/user.ts
+++ b/client/src/app/interfaces/user.ts
@@ -1,0 +1,16 @@
+import { Stock } from './stock';
+import { Transaction } from './transaction';
+
+export interface User {
+    firstName: string;
+    lastName: string;
+    username: string;
+    email: string;
+    profilePicture: string;
+    score: number;
+    availableBalance: number;
+    follows: User[];
+    followers: User[];
+    portfolio: Stock[];
+    transactionsMade: Transaction[];
+}
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -2298,6 +2298,47 @@
         "schema-utils": "^2.7.0"
       }
     },
+    "@ngrx/effects": {
+      "version": "10.0.1",
+      "resolved": "https://registry.npmjs.org/@ngrx/effects/-/effects-10.0.1.tgz",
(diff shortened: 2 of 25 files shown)


client/src/app/interfaces/user.ts
```diff
@@ -0,0 +1,16 @@
+import { Stock } from './stock';
+import { Transaction } from './transaction';
+
+export interface User {
+    firstName: string;
+    lastName: string;
+    username: string;
+    email: string;
+    profilePicture: string;
```
razine-bensari, line 9: I do not know yet if backend model has a profilePicture field. If it does not, how would we get it? I actually think we may need to use a CDN for user pictures if we cannot store them in Neo4J (because maybe neo4j does not support picture storage. We could store the link of the picture from the CDN to neo4j).

Is the string in that case the url of the picture?

razine-bensari (approved):
Super nice work!!! It made me think about user generated files that we need to store in the a content distributed netowrk (CDN). Either returned by the backend or fetched from the front end

client/src/app/interfaces/user.ts
Alessandro100: yes this is something we will have to discuss. We have two options
1. Give the user a preset amount of images to chose from (easiest)
2. Let the user upload a picture of themselves and store it on something like amazon S3 (better but more time consuming)

Whats going to be stored in NEO4j is just the url string of the image, not the actual image. When we get closer to those stories, we'll see our progress and decide which option to choose

PR #119, opened 2020-10-30
Iex

abdulmansour:
# Related Issue
- https://app.zenhub.com/workspaces/moneytree-board-5f5bcf1c7ea48f001c28c973/issues/dpigeon/money-tree/109 

# Proposed changes
- Able to make a sample call to the IEXCloud enpoint
- Implemented like the class diagram (Facade design pattern)

# Reviewer(s)
 - @razine-bensari


Alessandro100:
Hey so I briefly looked at this PR and it goes above its scope. Not only did you make the IEX Cloud connection, you added the controllers, models, decorators etc. This is not the end of the world, but just confirm that this stuff is done in this PR so that the other team doesn't overlap

razine-bensari:
This PR includes the commit from my PR that has already been merged, that is why it is showing 32 files. The commit history has been polluted. That is why

Issue #99, opened 2020-10-15
Setup backend architecture


Alessandro100:
As a developer I would like to setup the backend architecture so that it sets the base fo future work to go faster

Definition of setup:
Create all the class files, link them but the functionality of these functions found in the class diagram don't have to be implemented

**Acceptance Criteria**
- [x] Models are created 
- [x] Controllers are created
- [x] Services are created
- [x] The architecture diagrams are respected: ER diagram, Class diagram, component diagram
- [x] Integration Tests

razine-bensari:
A few thing that can be divided:

1) Add classes from achitecture diagrams and skeleton methods that we think we will need

2) Make sure the nodes and relationship are properly configured and persisting works well

3)Write the tests

4) Connect Facade with the external apis

5) Write integration tests for controllers

DPigeon:
We just had a meeting and talked about dividing the work. This story is the **base** architecture for the backend. Everyone working on the backend will be involved in this. It also has some deeper *sub-tasks*for the implementation and connection through the APIs and backend.

- Neo4j Sub-task: #110  (assigned to Razine)
- IEXCloud Sub-task: #109 (Assigned to Abdul)
- Alpaca Sub-task: #108 (Assigned to myself & Walter)

PR #123, opened 2020-11-02
Issue/#99 setup backend architecture

razine-bensari:
# Related Issue
- #99 

# Proposed changes
- Add architecture backend code

# Reviewer(s)
 - @DPigeon @abdulmansour @marwan011  @wltrf


DPigeon (approved):
Looks good! This is a good base setup for our backend. The CI seems to have an error. I had the same error about Decorators when building on the alpaca branch. It has something to do with the @ AllArgsConstructor if I remember.

backend/src/main/java/com/capstone/moneytree/model/decorator/OrderDecorator.java
```diff
@@ -0,0 +1,15 @@
+package com.capstone.moneytree.model.decorator;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.NoArgsConstructor;
+import net.jacobpeterson.domain.alpaca.order.Order;
+
+@Builder
```
DPigeon, line 8: Remove the @Builder and @NoArgsConstructor here as they are causing the error on the CI pipeline.

DPigeon (changes requested):

backend/src/main/java/com/capstone/moneytree/model/decorator/OrderDecorator.java
razine-bensari: Done!

backend/pom.xml
```diff
@@ -66,6 +66,17 @@
             <artifactId>spring-boot-autoconfigure</artifactId>
             <version>2.3.4.RELEASE</version>
         </dependency>
+        <dependency>
```
DPigeon, line 69: On line 145 (I can't write a comment on the line directly for some reasons):
```
<directory>${pom.basedir}/src/test/java</directory>
```
shouldn't this be 
```
<directory>${pom.basedir}/src/test/groovy</directory>
```
for tests instead?

backend/src/main/java/com/capstone/moneytree/facade/MarketInteractionsFacade.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * This facade abstracts the Alpaca api and exposes only the relevant
+ * methods for MoneyTree.
+ * */
+@Component
+public class MarketInteractionsFacade {
+
+   private final AlpacaAPI alpacaAPI = new AlpacaAPI();
+
+   @Autowired
```
DPigeon, line 17: The @ Autowired should be removed here to fix the CI.

DPigeon (changes requested):

backend/src/main/java/com/capstone/moneytree/model/OrderType.java
```diff
@@ -0,0 +1,8 @@
+package com.capstone.moneytree.model;
+
+public enum OrderType {
```
wltrf, line 3: We might want to change this in the future to split the "side" (buy/sell) and the "type" (market, limit, stop, ...). It would allow easier communication with the api and easier access to atomic order info. See https://alpaca.markets/docs/api-documentation/api-v2/orders/

backend/src/main/java/com/capstone/moneytree/model/TransactionStatus.java
```diff
@@ -0,0 +1,7 @@
+package com.capstone.moneytree.model;
+
+public enum TransactionStatus {
```
wltrf, line 3: This might need to be changed in the future to be more closely related to the api. See https://alpaca.markets/docs/trading-on-alpaca/orders/#order-lifecycle

backend/src/main/resources/application.properties
```diff
@@ -2,4 +2,17 @@ spring.profiles.active=dev
 
 spring.data.neo4j.uri=bolt://localhost:7687
```
wltrf, line 3: Note: We need to make sure we don't expose this port when deploying.

wltrf (approved):
Looks great, good job! I added some notes, but no changes are required.

backend/src/main/resources/application.properties
razine-bensari: yeah you are right : |

DPigeon:
I have fixed the Groovy dependency that was making the CI install backend phase failing. However now SonarCloud is saying that we have no code coverage on the backend. @wltrf let me know when you fix this. We need it for this PR on the backend.

DPigeon (approved):

Issue #122, opened 2020-11-01
Response Error Handling


DPigeon:
As a user, I want to receive a good detailed response when I use the app so that responses are handled properly.

After working a bit on the backend, we realized that the error handling of responses was something to consider. Indeed, the team is using the **ResponseEntity** to handle the responses on the backend. Unfortunately, this is not enough for us to handle errors properly on the responses we sent to the frontend. Therefore, a method must be investigated to properly handle errors on our responses created onto the backend. The following explains what could be investigated:
The ResponseEntity can be used for that with its builder when catching exceptions:
```
new ResponseEntity
        .status(int status)
        .body(String message);
```

Or we can use the **Optional** API with the ResponseEntity as follows (Optional is composed of any response needed and it can easily be retrieved as any responses):
```
new ResponseEntity.of(Optional optionalObject);
```

**Acceptance of criteria:**
- [ ] Investigated a good way to handle responses
- [ ] A new class is created to be able to reuse the error handling process

DPigeon:
@razine-bensari and I just had a meeting about this story. We'll be using a MoneyTreeError class with an enum ErrorMessage to make the code cleaner. ResponseEntity and optionals are also going to be used.

Issue #108, opened 2020-10-19
Setup Alpaca


Alessandro100:
As a developer I would like to setup Alpaca in the backend so future developers can easily work on top of it

Note:
- For the API keys, reach out to me and I'll create them for you - Alessandro

**Acceptance Criteria**
- The Alpaca SDK is setup
- We are able to make a successful test call to an Alpaca endpoint
- It is setup as shown in the software class diagram (Facade)

DPigeon:
For now, Alpaca has been setup properly with my Alpaca`s secret keys for testing. Three different endpoints were to test locally. Everything is setup as shown in the class diagram. However, we still have to setup the Order Decorators. I am still unsure if we want to write the basic methods that will be used for the Order Decorators now or if we just want a simple Order Decorator setup like what was done by @razine-bensari and @amansourcoveo already in the iex & setup backend architecture branch.

razine-bensari:
We can leave order decorator methods for now, because for the first set of features/ release 1 we will not be making orders just yet.

Once we work on the feature that needs us to create order, we will implement the methods related to orders decorators.

WDYT?

DPigeon:
> We can leave order decorator methods for now, because for the first set of features/ release 1 we will not be making orders just yet.
> 
> Once we work on the feature that needs us to create order, we will implement the methods related to orders decorators.
> 
> WDYT?

Alright I agree. They are setup for now anyway and the files are made.

PR #124, opened 2020-11-04
#108: Setup Alpaca

DPigeon:
# Related Issue
- #108

# Proposed changes
- 3 exposed endpoints in the AlpacaController
- MarketInteractionsFacade builds the proper needed Alpaca objects
- 8 Integration tests on three different endpoints

# Additional Info
- Note: We cannot really test many exceptions or errors in the integration tests since we need a better robust error handling (#122)  and validation strategy before doing so.  
- I also left some TODOs for later. Let me know if something is not clear.

# Checklist (if applicable)
- [x] Integration Tests

# Reviewer(s)
 - @razine-bensari @abdulmansour @wltrf @Alessandro100


backend/src/main/java/com/capstone/moneytree/controller/AlpacaController.java
```diff
@@ -0,0 +1,112 @@
...
+        this.marketInteractionsFacade = marketInteractionsFacade;
+    }
+
+    /**
+     * Get an Alpaca account endpoint.
+     * @return ResponseEntity of Account.
+     */
+    @GetMapping("/account")
+    public ResponseEntity<Account> getAccount() {
+        Account account = marketInteractionsFacade.getAccount();
```
razine-bensari, line 40: is the getAccount user level or application level? (we get the account of the user?
)

DPigeon: Yeah we get all of an Alpaca user. https://alpaca.markets/docs/api-documentation/api-v2/account/

razine-bensari: But how do we specify the user?

razine-bensari:
REALLY nice work!, we need to add Java doc in already written and merged code. Thks a lot for adding java doc.

razine-bensari (approved):

backend/src/main/java/com/capstone/moneytree/controller/AlpacaController.java
DPigeon: Alpaca will work with an OAuth ID later when we configure authentication. We'll get the account of a user with its secret key using the auth secret of the properties. Let's discuss about it more next sprint.

Issue #34, opened 2020-09-17
Display Stock Price


abdulmansour:
As a user I want to be able to see the current price of a stock so I see if I want to buy or sell

Note:
- Don't worry about the stock exchange and the risk for this story
- Don't worry about displaying the right information in this story, getting the data will be done in another story

**Acceptance Criteria**
- Display the current price of the stock
- Display the daily change of the stock
- Display the title of the stock
- Display the logo of the stock
- The ticker of the stock (AC) must route to the correct stock page ex: money-tree/stock-details/AC (this is an example)
- If the daily change is up, the text must be green
- If the daily change is down, the text must be red
- Must have unit tests for the color of the text

![Screen Shot 2020-10-16 at 6 25 44 AM](https://user-images.githubusercontent.com/18631060/96247687-777b6600-0f78-11eb-9725-6cbf5086632a.png)

PR #126, opened 2020-11-05
Feature#34/dispaly stock price/new tests

EspressoCode:
# Related Issue
- #34

# Proposed changes
- adds two new e2e tests for arrow up and arrow down
- adds the case where the stock change is 0, we don't show any arrow for that case, and the text color will be black

# Additional Info
I did not directly commit to our previous branch so that you can review changes and if you did not approve we will be good to approve your PR and merge to dev. I just wanted to practice testing and add more possible cases, I checked with positive, negative and 0 values for stock change and all of them passed.

### DIFF
--- a/client/src/app/components/stock-detail-header/stock-detail-header.component.ts
+++ b/client/src/app/components/stock-detail-header/stock-detail-header.component.ts
@@ -12,11 +12,19 @@ export class StockDetailHeaderComponent implements OnInit{
 
   ngOnInit(): void {}
 
-  stockChangeColor() {
-    return this.stockInfo.stockChange < 0 ? 'negative-change' : 'positive-change';
+  get stockValue() {
+    return this.stockInfo ? this.stockInfo.stockValue : 0;
   }
+
+  stockChangeColor() {
+    return this.stockInfo && this.stockInfo.stockChange < 0 ? 'negative-change' : 'positive-change';
+  }
+
   stockInfoFormatter() {
-    let sign = this.stockInfo.stockChange < 0 ? '-' : '+';
-    return sign + this.stockInfo.stockChange + "("+ this.stockInfo.stockChangePercent+"%)";
+    if(this.stockInfo) {
+      let sign = this.stockInfo.stockChange < 0 ? '-' : '+';
+      return sign + this.stockInfo.stockChange + "("+ this.stockInfo.stockChangePercent+"%)";
+    }
+    return '';
   }
 }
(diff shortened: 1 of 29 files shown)


client/src/app/components/stock-detail-header/stock-detail-header.component.ts
```diff
@@ -15,16 +15,22 @@ export class StockDetailHeaderComponent {
   }
 
   stockChangeColor() {
-    if(this.stockInfo) {
-      return this.stockInfo.stockChange < 0 ? 'negative-change' : 'positive-change';
+    if (this.stockInfo) {
```
Alessandro100, line 18: This function can be cleaner by doing

`if (this.stockInfo && this.stockInfo.stockChange !==0)`

Since both these conditions lead to returning ' ' this is valid

client/src/app/components/stock-detail-header/stock-detail-header.component.ts
```diff
@@ -15,16 +15,22 @@ export class StockDetailHeaderComponent {
...
+      const changeValue = this.stockInfo.stockChange;
+      if (changeValue === 0) {
+        return '';
+      }
```
Alessandro100, line 22: With the above implemented, we can remove this

client/src/app/components/stock-detail-header/stock-detail-header.component.ts
```diff
@@ -15,16 +15,22 @@ export class StockDetailHeaderComponent {
...
     }
     return '';
   }
 
   stockInfoFormatter() {
-    if(this.stockInfo) {
-      let sign = this.stockInfo.stockChange < 0 ? '' : '+';
-      return (sign + this.stockInfo.stockChange + "("+ this.stockInfo.stockChangePercent+"%)");
+    if (this.stockInfo) {
+      let sign = this.stockInfo.stockChange <= 0 ? '' : '+';
```
Alessandro100, line 32: Since we have another logic branch, a new unit test case will need to be written

Alessandro100 (changes requested):
Thank you for taking the initiative to add onto the e2e test and expand the logic. You don’t need to create a separate PR, you could have done a normal commit and I can review it there. That being said, I think you might have branched off the wrong branch because in this PR I see 29 changed files (which is most of the work we did on the main branch). To simplify our lives, I would suggest redoing the changes (with my comments) directly on the main branch. If you want to do this together, I'll be free to program with you today, otherwise I'll be free for the review

EspressoCode:
I added new changes in a commit on the main branch, so I closed this pr.

Issue #109, opened 2020-10-19
Setup IEXCloud


Alessandro100:
As a developer I would like to setup IEXCloud so future developers can work with it

Note:
- I have an API key for this, let me know and I'll give it to you

**Acceptance Criteria**
- We are able to make a sample call to the IEXCloud enpoint
- It is implemented like the class diagram (Facade design pattern)

PR #127, opened 2020-11-05
Issue/#109 setup iexcloud

razine-bensari:
# Related Issue
- #109 

# Proposed changes
- Setup IEXCloud for stock user story

# Checklist (if applicable)
- [ ] Tests not yet. Will be updated with PR commit

# Reviewer(s)
 - @Alessandro100 @razine-bensari  @abdulmansour @DPigeon

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/controller/StockController.java
+++ b/backend/src/main/java/com/capstone/moneytree/controller/StockController.java
@@ -8,39 +8,76 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.*;
 
 import com.capstone.moneytree.service.api.StockService;
-import pl.zankowski.iextrading4j.api.stocks.Quote;
+import pl.zankowski.iextrading4j.api.stocks.*;
 import pl.zankowski.iextrading4j.api.stocks.v1.BatchStocks;
+import pl.zankowski.iextrading4j.api.stocks.v1.KeyStats;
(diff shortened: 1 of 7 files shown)


backend/src/main/java/com/capstone/moneytree/controller/StockController.java
```diff
@@ -1,23 +1,83 @@
...
-      this.stockService = stockService;
+   /**
+    * Obtain json object containing company quote, info, news, keystats and logo
+    *
+    * @param symbol
+    * @return batchStocks
+    */
+   @GetMapping("/batch/{symbol}")
+   public ResponseEntity<BatchStocks> getBatchStocksBySymbol(@Valid @PathVariable(name = "symbol") String symbol) {
+      BatchStocks batchStocks = stockMarketDataService.getBatchStocksBySymbol(symbol);
```
razine-bensari, line 37: Is there a way to convert this BatchStock from IEXCloud library to the Stock class that we actually persist in NEO4J.

So front end can simply receive our Stock object. WDYT?

razine-bensari: Actually, now that I think about it. There is no need. Stock class is the stock bought by the user. BatchStock is what the front end needs with all info about stocks

DPigeon (approved):
Nice work! Approved.

Alessandro100 (approved):
Looks like good work, the one thing to investigate is how sonarcloud works with our coverage as it says 0 but we do have tests. This can be done in a later commit

PR #129, opened 2020-11-06
Feature/get stock price information #111

Alessandro100:
# Related Issue
- [111](https://github.com/DPigeon/Money-Tree/issues/111)

# Proposed changes
- Connected the backend endpoint to the frontend
- Display real stock data
# Additional Info
- To run and test this, the backend and database must be running
- When testing locally, you'll need this chrome extension turned on https://chrome.google.com/webstore/detail/moesif-origin-cors-change/digfbfaphojjndkpccljibejjbppifbc
- TODO: From the backend, enable CORS access to localhost. Discuss the way we handle data from the backend

# Checklist (if applicable)
- [X] Tests
- [X] Documentation

# Screenshots (if applicable)
Screen Shot 2020-11-06 at 8 48 46 AM
Screen Shot 2020-11-06 at 8 48 33 AM

# Reviewer(s)
 - @arthur-del 
 - @marwan011
 - @EspressoCode 
 - @DPigeon

PR #130, opened 2020-11-06
Issue/#103: Investigate how to deploy Neo4J on Docker container

### LINKED ISSUE #103: Investigate how to deploy Neo4J on a docker container
As a user, I want to access my personal data on the app so that I can see my information properly.

When deploying to the 2 environments, we need to find a way to interact with the database. To achieve that, we must investigate a good strategy to persist our database data in production. On the other hand, the UAT environment can have test data and can be deleted at any time since it is a "test" environment. With the current CD pipeline script, a good solution must be investigated for both environments.

**Acceptance of Criteria:**

- [x] Figure out if we can simply incorporate a docker container for the neo4j db within the pipeline.

- [x] Figure out if when we do `docker image prune` it does not remove the db data.

- [x] Figure out if when we do `docker system prune` it does not remove the db data.

- [x] Figure out if when we stop a container and redeploy, or rebuild an image, the db data is not removed.

wltrf:
# Related Issue
- closes #103 

# Proposed changes
- Run neo4j in a container instead of separately

# Additional Info
- This is specific to deployment only. Instead of running using docker run, we will use docker-compose up.
- Action item:
- [ ] Change the deployment script in the vm to use docker-compose up

# Checklist (if applicable)
- [x] Tests N/A
- [x] Documentation N/A Documented in the issue

# Screenshots (if applicable)
N/A

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# Video or links (if applicable)
 - url

# Reviewer(s)
 - @DPigeon

### DIFF
--- a/backend/docker-compose.yml
+++ b/backend/docker-compose.yml
@@ -1,0 +1,19 @@
+services:
+  money-tree-application:
+    build: .
+    restart: on-failure
+    ports:
+      - 8080:8080
+    links:
+      - neo4j
+    depends_on:
+      - neo4j
+  neo4j:
+    image: neo4j
+    restart: unless-stopped
+    ports:
+      - 7474:7474
+      - 7687:7687
+    volumes:
+      - /home/neo4j/data:/data
+      - /home/neo4j/logs:/logs


backend/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    restart: on-failure
+    ports:
+      - 8080:8080
+    links:
+      - neo4j
+    depends_on:
+      - neo4j
+  neo4j:
+    image: neo4j
+    restart: unless-stopped
```
DPigeon, line 13: So this line means that the database should be on to always persist data and if it is stopped, we restart it right?

DPigeon (commented):
Nice job! I left a question about the management of the DB.

backend/docker-compose.yml
wltrf: It means if the service stops for any reason (docker daemon restarts, vm power cut, ...), it will restart the database service, unless we stopped it ourselves. In our case, this line is optional, and I don't see when it use it, but I put it there for extra robustness in case i missed something.
See documentation [here](https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy)

wltrf:
The pipeline fails because the sonarcloud fails for backend. It can be ignored since no code was changed and probably comes from another pr.

backend/docker-compose.yml
DPigeon: Ok I see! Good stuff.

DPigeon:
So we only need to change the deployment script into the VM. The plan is to modify it so that we run the frontend with a simple docker image and the backend & database with the docker-compose file we have in this PR?

wltrf:
Yes

DPigeon (approved):
Approved! Good work.

DPigeon:
> The pipeline fails because the sonarcloud fails for backend. It can be ignored since no code was changed and probably comes from another pr.

Actually the pipeline fails because we exceeded our quota for IEXCloud. But you are right since no code changed in the repo we can skip this. Based on a newer PR #129, we are sure that the CI tests passes as we have changed the API keys for IEXCloud allowing us to have more requests calls available. 

I approved this PR. Let's now merge it and skip the CI for this one.

PR #148, opened 2020-11-11
Issue #122: error handling

razine-bensari:
# Related Issue
- #122 

# Additional Info
- Added a generic MoneyTreeError class to wrap arround in our API with ErrorCode. Changed the IEXCloud library to take
the one I forked from github because I added a new endpoint for sandbox and released it under tag 3.3.5 so we can use it locally
- changed IEXCloud apis keys to use Sandbox environment

# Checklist (if applicable)
- [x] Tests
- [ ] Documentation

# Reviewer(s)
 - @abdulmansour @Alessandro100 @DPigeon


backend/pom.xml
```diff
@@ -163,4 +168,10 @@
             </plugin>
         </plugins>
     </build>
+    <repositories>
+        <repository>
+            <id>jitpack.io</id>
```
DPigeon, line 173: Good idea! I would've have never thought about this. Nice!

backend/src/main/java/com/capstone/moneytree/utils/MoneyTreeError.java
```diff
@@ -0,0 +1,34 @@
+package com.capstone.moneytree.utils;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.NoArgsConstructor;
+
+@Builder
```
DPigeon, line 7: Can you also make a method that creates a MoneyTreeError object? So that we dont have to write redundant code everytime we create a MoneyTreeError. Lombok probably already does that though.

```
MoneyTreeError createError(String message) {
      return MoneyTreeError.builder()
                       .setMessage(message);
}
```

DPigeon (changes requested):
Nice work! What we talked about was quickly implemented :).
I left a comment about the MoneyTreeError.

Good work!

backend/src/main/java/com/capstone/moneytree/utils/MoneyTreeError.java
```diff
@@ -0,0 +1,34 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+public class MoneyTreeError {
+
+    private MoneyTreeErrorCode errorCode;
+    private String message;
```
DPigeon, line 13: Here we have two messages. MoneyTreeErrorCode is a message itself. I think we should remove the message parameter.

DPigeon (changes requested):

backend/src/main/java/com/capstone/moneytree/utils/MoneyTreeError.java
DPigeon: Also, I just thought about this and we're still missing a way to generate a status, message and body for the response all at the same time.. Remember that in the end, we'll return a ResponseEntity (with status needed and body) to the frontend as a final object when creating a user is completed for example.
So, I think we need a better way to handle this. What I am thinking right now is having a MoneyTreeBody class that will have:
- Error Code Message
- Body of any Object

Because we'll always be returning this:
`
return ResponseEntity
         .status(status)
         .body(moneyTreeBody)
`

And that body will have to have an object of any kind and a message for the frontend to know whats going on.

backend/pom.xml
```diff
@@ -62,11 +62,11 @@
...
-        <dependency>
-            <groupId>pl.zankowski</groupId>
-            <artifactId>iextrading4j-all</artifactId>
-            <version>3.3.4</version>
-        </dependency>
+<!--        <dependency>-->
+<!--            <groupId>pl.zankowski</groupId>-->
+<!--            <artifactId>iextrading4j-all</artifactId>-->
+<!--            <version>3.3.4</version>-->
+<!--        </dependency>-->
```
Alessandro100, line 69: If we aren't using this code, we should delete it

Alessandro100 (changes requested):
Why did we include error handling in this PR? It's not the end of the world, but we should keep PR's focused to the main subject which in this case was 'Changing IEXCloud keys'

razine-bensari:
This story #122 is about error handling, not API keys. I added the new api keys since the previous ones were not correct.

backend/src/main/java/com/capstone/moneytree/handler/RestExceptionHandler.java
```diff
@@ -0,0 +1,35 @@
...
+@ControllerAdvice
+public class RestExceptionHandler extends ResponseEntityExceptionHandler {
+
+   private ResponseEntity<MoneyTreeError> buildResponseEntity(MoneyTreeError apiError) {
+      return new ResponseEntity<>(apiError, apiError.getStatus());
+   }
+
+   //other exception handlers below
+   @ExceptionHandler(EntityNotFoundException.class)
+   protected ResponseEntity<MoneyTreeError> handleEntityNotFound(
```
DPigeon, line 25: Nice! So if we validate an endpoint for example and see that there are no errors, then we would return a ResponseEntity<Object> like before. However, if we have errors, then we would create a ResponseEntity<MoneyTreeError> from the different handles we have here. Then Spring would automatically understand return the right entity?

razine-bensari: Yeah so, this handler class will handle all exception we indicate. If we throw an exception while processing a request like user not found exception, spring will go to this handler and see which handler handles that specific exception. Then it will return the response entity with our error object.

So there is no need to have any kind of wrapper or anything like that

DPigeon (approved):
Nice work! Approved.

DPigeon:
I pushed some of the Alpaca exceptions that we could get such as NullPointerException, IllegalArgumentException and more. Refactored a little bit the endpoints I did on Alpaca. Tested and works well! We definetely handle error responses better now and it is very simple. Good job!

Alessandro100 (approved):
Ah ok, the previous title of the PR was 'IEXCloud change keys' so I was a little confused, but this makes sense, good job

Issue #143, opened 2020-11-09
Display stock industry


Alessandro100:
As a user I would like to view the stock industry to know what I'm investing in

**Acceptance Criteria**
- Unit tests
- System tests
- When a user goes on the stock-detail page, they can view the stock industry

![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/8eb8b99d-19dc-45a4-b847-c5b7e9373d27)

Issue #39, opened 2020-09-17
Display Stock Industry


abdulmansour:
As a user I want to know the stock belongs to which industry ex: mining, tech, pharma...

**Acceptance Criteria**
- Able to see which industry the stock belongs to
- e2e test is written

![Screen Shot 2020-11-02 at 2 20 20 PM](https://user-images.githubusercontent.com/18631060/97909657-9880dc80-1d16-11eb-9e96-f0a6d0c460eb.png)

lindsaybangs:
Accepted after demonstration on Friday, November 27

PR #150, opened 2020-11-12
Feature/# 143 display stock industry 

marwan011:
# Related Issue
https://github.com/DPigeon/Money-Tree/issues/143

# Proposed changes
- add industry field as a mat-chip
- change background color of the pages 

# Checklist (if applicable)
- [X] Tests 
# Reviewer(s)
 - @Alessandro100

### DIFF
--- a/client/src/app/app.component.scss
+++ b/client/src/app/app.component.scss
@@ -1,4 +1,4 @@
-.allPages {
+.all-pages {
     background-color: #F8F8FB;
     width: 100%;
     height: 100%;
--- a/client/src/app/app.component.scss
+++ b/client/src/app/app.component.scss
@@ -1,0 +1,5 @@
+.allPages {
+    background-color: #F8F8FB;
+    width: 100%;
+    height: 100%;
+}
--- a/client/src/app/app.module.ts
+++ b/client/src/app/app.module.ts
@@ -16,6 +16,7 @@ import { StockDetailHeaderComponent } from './components/stock-detail-header/sto
 import { MatCardModule } from '@angular/material/card';
 import { FlexLayoutModule } from '@angular/flex-layout';
 import { MatIconModule } from '@angular/material/icon';
+import {MatChipsModule} from '@angular/material/chips';
 
 
 @NgModule({
@@ -34,7 +35,8 @@ import { MatIconModule } from '@angular/material/icon';
     EffectsModule.forRoot([Effects]),
     MatCardModule,
     FlexLayoutModule,
-    MatIconModule
+    MatIconModule,
+    MatChipsModule
   ],
   providers: [],
   bootstrap: [AppComponent]
--- a/client/src/app/app.module.ts
+++ b/client/src/app/app.module.ts
@@ -16,7 +16,7 @@ import { StockDetailHeaderComponent } from './components/stock-detail-header/sto
 import { MatCardModule } from '@angular/material/card';
 import { FlexLayoutModule } from '@angular/flex-layout';
 import { MatIconModule } from '@angular/material/icon';
-import {MatChipsModule} from '@angular/material/chips';
+import { MatChipsModule } from '@angular/material/chips';
 
 
 @NgModule({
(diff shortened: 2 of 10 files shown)


client/src/app/app.module.ts
```diff
@@ -16,6 +16,7 @@ import { StockDetailHeaderComponent } from './components/stock-detail-header/sto
 import { MatCardModule } from '@angular/material/card';
 import { FlexLayoutModule } from '@angular/flex-layout';
 import { MatIconModule } from '@angular/material/icon';
+import {MatChipsModule} from '@angular/material/chips';
```
Alessandro100, line 19: Small: {MatChipsModule} should be { MatChipsModule }  (add spaces

client/src/app/app.component.scss
```diff
@@ -0,0 +1,5 @@
+.allPages {
```
Alessandro100, line 1: scss naming convention is 'all-pages'

client/src/app/components/stock-detail-header/stock-detail-header.component.html
```diff
@@ -1,6 +1,12 @@
 <div class="company-title">
+    <span>
```
Alessandro100, line 2: this should be a <div> and should have a class name to it, its bad practice to style a span / div with nothing

div = a block of elements
span = an inline element

client/src/app/components/stock-detail-header/stock-detail-header.component.html
```diff
@@ -1,6 +1,12 @@
 <div class="company-title">
+    <span> 
     <div class="company-logo" [style.background-image]='companyLogo'></div>
-    <h1><span class="ticker">{{ companySymbol }}</span> {{ companyName }}</h1>
+        <h1><span class="ticker">{{ companySymbol }}</span> {{ companyName }}</h1>
+    </span>   
+        <mat-chip-list>
+            <!-- <mat-chip color="primary">Open/close market</mat-chip> -->
```
Alessandro100, line 7: this comment on line 7 should be removed

client/src/app/components/stock-detail-header/stock-detail-header.component.scss
```diff
@@ -4,26 +4,36 @@
     display: flex;
     align-items: center;
     margin-bottom: 20px;
+    justify-content: space-between;
+    align-items: center;
+    
+    span {
```
Alessandro100, line 10: As mentioned above, we should not use a spam, and should add a class name to it instead

client/src/app/components/stock-detail-header/stock-detail-header.component.ts
```diff
@@ -25,6 +25,9 @@ export class StockDetailHeaderComponent {
   get stockValue(): string {
     return this.stockInfo ? this.stockInfo.stockValue.toFixed(2) : '';
   }
+  get stockIndustry(): string {
+    return this.stockInfo ? this.stockInfo.industry : '';
+  }
```
Alessandro100, line 30: A test should be written for this function. Yes I know it's small, but it has logic so we must test it

Alessandro100 (changes requested):
I left some changes, some are minor, some are more in-depth. Another thing I noticed is that you forgot to write a system test for this feature. Because we have an acceptance test for this story, we need to have a system test for it

https://github.com/DPigeon/Money-Tree/issues/39

Let me know if you have any questions / stuff I can help you out with

client/src/app/components/stock-detail-header/stock-detail-header.component.scss
```diff
@@ -4,26 +4,33 @@
     display: flex;
     align-items: center;
     margin-bottom: 20px;
+    justify-content: space-between;
+    align-items: center;
+    
+    div {
```
Alessandro100, line 10: we should not have this as a div but rather a class name. It is bad practice to style a naked div.

.stock-headline type thing

client/src/app/components/stock-detail-header/stock-detail-header.component.html
```diff
@@ -1,6 +1,11 @@
 <div class="company-title">
+    <div>
```
Alessandro100, line 2: add a class name to this div and style it with the class name

Alessandro100 (changes requested):
I left a small comment regarding class name, but you also forgot to write a system test for this story!

Alessandro100 (approved):
nice changes!

Issue #141, opened 2020-11-09
IEX Cloud API keys


Alessandro100:
As a developer I want to optimize IEX Cloud usage to better scale

Todo: Discuss best IEX practices

**Acceptance Criteria**
- Use IEX sandbox environment in tests (or mock them)
- Analyze how we can reduce our IEX usage
- Remove 'news' and 'key statistics'

razine-bensari:
`This is what you need to add to you environment variables

#For MoneyTree
export ALPACA_KEY_ID="PKZJK3IXQ4JZAWCBDSRX"
export ALPACA_SECRET_ID="SA5sYIYR58AukN8NKpdGkQZSCvOCoTXYgyjuaoUe"
export IEXCLOUD_PUBLISHABLE_TOKEN_SANDBOX="Tpk_41cea3476f55437aa0a7ff89162ea9c4"
export IEXCLOUD_SECRET_TOKEN_SANDBOX="Tsk_072d4c09f38148a9b68b748b38fad53c"

export IEXCLOUD_PUBLISHABLE_TOKEN_PROD=""
export IEXCLOUD_SECRET_TOKEN_PROD=""

export NEO4J_USERNAME="neo4j"
export NEO4J_PASSWORD="stonecap"


`

PR #151, opened 2020-11-12
Issue/#141 iex api keys

razine-bensari:
# Related Issue
- #141 

# Proposed changes
- The proposed change is to put all secret api keys and sensitive info into our environment variables. Need to update the VM in the CI/CD. I also fixed the api path for the controllers

# Reviewer(s)
 - @DPigeon @Alessandro100 @abdulmansour


razine-bensari:
Here are the environment variables to add for all of us:

#For MoneyTree
export ALPACA_KEY_ID="PKZJK3IXQ4JZAWCBDSRX"
export ALPACA_SECRET_ID="SA5sYIYR58AukN8NKpdGkQZSCvOCoTXYgyjuaoUe"
export IEXCLOUD_PUBLISHABLE_TOKEN_SANDBOX="Tpk_41cea3476f55437aa0a7ff89162ea9c4"
export IEXCLOUD_SECRET_TOKEN_SANDBOX="Tsk_072d4c09f38148a9b68b748b38fad53c"

export IEXCLOUD_PUBLISHABLE_TOKEN_PROD=""
export IEXCLOUD_SECRET_TOKEN_PROD=""

export NEO4J_USERNAME="neo4j"
export NEO4J_PASSWORD="stonecap"

DPigeon:
> Here are the environment variables to add for all of us:
> 
> #For MoneyTree
> export ALPACA_KEY_ID="PKZJK3IXQ4JZAWCBDSRX"
> export ALPACA_SECRET_ID="SA5sYIYR58AukN8NKpdGkQZSCvOCoTXYgyjuaoUe"
> export IEXCLOUD_PUBLISHABLE_TOKEN_SANDBOX="Tpk_41cea3476f55437aa0a7ff89162ea9c4"
> export IEXCLOUD_SECRET_TOKEN_SANDBOX="Tsk_072d4c09f38148a9b68b748b38fad53c"
> 
> export IEXCLOUD_PUBLISHABLE_TOKEN_PROD=""
> export IEXCLOUD_SECRET_TOKEN_PROD=""
> 
> export NEO4J_USERNAME="neo4j"
> export NEO4J_PASSWORD="stonecap"

I've added all of these into secrets into our repo. You can access them on the GitHub Actions workflow with:

```
{{ secrets.ALPACA_KEY_ID }}
{{ secrets.ALPACA_SECRET_ID }}
{{ secrets.IEXCLOUD_PUBLISHABLE_TOKEN_SANDBOX }}
{{ secrets.IEXCLOUD_SECRET_TOKEN_SANDBOX }}

{{ secrets.NEO4J_USERNAME }}
{{ secrets.NEO4J_PASSWORD }}
```

Note that nothing was specified for the IEXCLOUD_PRODs tokens so we can add them later when we have them.

backend/src/main/resources/application.properties
```diff
@@ -1,20 +1,23 @@
...
+spring.data.neo4j.username=${NEO4J_USERNAME}
+spring.data.neo4j.password=${NEO4J_PASSWORD}
 
 # Alpaca Related properties
 # Client ID Auth: 198903d0d2f523a25e4dce65837bbf0d
 # Client Secret Auth: d756cb7ae52ad380d16aec004a38b00f04747cf2
 alpaca.api.version=v2
-alpaca.key.id=PKZJK3IXQ4JZAWCBDSRX
-alpaca.secret=SA5sYIYR58AukN8NKpdGkQZSCvOCoTXYgyjuaoUe
+alpaca.key.id=${ALPACA_KEY_ID}
```
DPigeon, line 16: You cannot use the secrets defined on our repo from here. You have to inject them from the CI/CD workflow. I believe there may be a way to achieve this but I haven't looked into it yet for Spring.

DPigeon (changes requested):

backend/src/main/resources/application.properties
razine-bensari: The values are injected if I have them present within the environment variable of my computer. I tested it locally it works well

Even in the tests

razine-bensari: Maybe I misunderstood your comment?

DPigeon: So GitHub Actions secrets can only be used within the DevOps pipeline workflow. It doesn't see the secrets at all on your local machine. It has to go through the CI/CD to be able to use the secret token. So I think the best solution here would be to separate everything into 3 environments (local, CICD and prod).
- If we test the endpoints locally and we need to use some API keys, we could use the sandbox keys. (need to figure out a way to hide the key though)
- If we test on the CICD, then we can use the secrets easily with 
```
${{ secrets.TOKEN_NAME }}
```
- If we deploy, we need to add all needed keys in the classpath of our JAR file before packaging it in the CD. Again, it can be easily achieved with something like 
```
jar -cp ALPACA_KEY=${{ secrets.ALPACA }} etc...
``` 

With those 3 cases, we need to detect which env we are in and apply whatever case above.

backend/src/main/resources/application.properties
```diff
@@ -1,20 +1,23 @@
...
+alpaca.secret=${ALPACA_SECRET_ID}
 alpaca.base.api.url=https://paper-api.alpaca.markets
 alpaca.base.data.url=https://data.alpaca.markets
 
 # IEXCloud related properties
-IEXCloud.publishable.token=Tpk_41cea3476f55437aa0a7ff89162ea9c4
-IEXCloud.secret.token=Tsk_072d4c09f38148a9b68b748b38fad53c
\ No newline at end of file
+IEXCloud.publishable.token=${IEXCLOUD_PUBLISHABLE_TOKEN_SANDBOX}
+IEXCloud.secret.token=${IEXCLOUD_SECRET_TOKEN_SANDBOX}
```
wltrf, line 23: If we append _SANDBOX to the env variables for iexcloud here, it would mean that we also use the sandbox keys for prod. I see 2 solutions to this:

1. We rename the environment variable by removing the sandbox suffix and use the appropriate key in the different environment
e.g. 
- local env: export IEXCLOUD_*_TOKEN = sandbox key
- ci env: export IEXCLOUD_*_TOKEN = sandbox key
- dev env: export IEXCLOUD_*_TOKEN = sandbox key
- prod env: export IEXCLOUD_*_TOKEN = prod key
 
OR

2. We have both (or just one) of prod and sandbox keys on all environment, and we decide which one we use inside the application by making use of the spring profile property (dev or prod)

backend/src/main/resources/application.properties
wltrf: I think there is some confusion here so I'll try to clarify. What we are trying to achieve here is hide the keys / make the application packaging independent of the keys. In order to do so, we make use of environment variables in the different environments where the application will run. The application is going to pick up these variables when running. The only reason we also store them in github secrets, is to be able to expose them as environment variables in the ci environment without exposing them in the ci.yml file.

wltrf (changes requested):
Refactor the following code smells:
https://sonarcloud.io/project/issues?id=money-tree_BACKEND&pullRequest=151&resolved=false&types=CODE_SMELL

DPigeon (commented):
Good changes! Now the only thing that is left to decide is walter`s comment about the IEXCloud local, ci, dev and prod keys.
By the way, before I forget: in the ci.yml file, could you change the line 109:
```
run: bash ./mvnw install
```
to
```
run: bash ./mvnw install -Dmaven.test.skip=true
```
and line 117:
```
run: bash ./mvnw package
```
to
```
run: bash ./mvnw package -Dmaven.test.skip=true
```
Please? I know this is not related to the PR but it will make our CI a bit faster by skipping the tests since we're already doing the tests in the Backend Tests stage.

DPigeon (commented):
I think that when we deploy, we also need to inject the keys into the package classpath in order to have them in uat and prod. I think we're missing that part.

wltrf (approved):

wltrf:
> I think that when we deploy, we also need to inject the keys into the package classpath in order to have them in uat and prod. I think we're missing that part.

I wouldn't package the application with the keys since:
1. If we want to change the keys, we need to repackage the application and
2. If anyone has the packaged application/image, one has also our keys which is a security liability

Simply having the environment variables on the dev and prod vms should suffice. Correct me if I'm missing something here.

DPigeon:
> > I think that when we deploy, we also need to inject the keys into the package classpath in order to have them in uat and prod. I think we're missing that part.
> 
> I wouldn't package the application with the keys since:
> 1. If we want to change the keys, we need to repackage the application and
> 2. If anyone has the packaged application/image, one has also our keys which is a security liability
> 
> Simply having the environment variables on the dev and prod vms should suffice. Correct me if I'm missing something here.

Very good point! Then we need to inject the keys onto the deployment script. However I am not sure if that is possible. We can probably append some variables into the Dockerfile RUN line of the backend where we inject the keys.

backend/src/main/resources/application.properties
razine-bensari: I just saw this comment, Ill update the PR

wltrf:
Since the dev,ci, and uat environments all have the same configuration (sandbox keys), maybe we can omit the uat and ci application.properties files and only use dev/prod as per my original comment.

razine-bensari:
> Since the dev,ci, and uat environments all have the same configuration (sandbox keys), maybe we can omit the uat and ci application.properties files and only use dev/prod as per my original comment.

My bad, I am getting confused with what is needed for the CI. I updated the PR with only two properties file. MAybe we can discuss with @DPigeon  how we can properly manage the keys for both the CI and the VMs in which we deploy the artifacts

DPigeon:
> > Since the dev,ci, and uat environments all have the same configuration (sandbox keys), maybe we can omit the uat and ci application.properties files and only use dev/prod as per my original comment.
> 
> My bad @wltrf , I am getting confused with what is needed for the CI. I updated the PR with only two properties file. MAybe we can discuss with @DPigeon how we can properly manage the keys for both the CI and the VMs in which we deploy the artifacts.

Yes we will have a discussion. I'll approve this now since we're tackling the management of the keys in the deployment in another story.

DPigeon (approved):
Approved! Good job.

wltrf (approved):

PR #152, opened 2020-11-13
issue/#149 backend coverage

### LINKED ISSUE #149: Sonarcloud backend is not showing coverage
**Describe the bug**
In sonarcloud, it shows that the backend has 0% coverage which is not the case. It does not seem to pick up clover

**To Reproduce**
1. Go to [sonarcloud](https://sonarcloud.io/dashboard?id=money-tree_BACKEND)
2. You'll see that the coverage is 0

**Expected behavior**
We expect the coverage to reflect the values given in clover

Bug reported by @Alessandro100

wltrf:
# Related Issue
- closes #149 

# Proposed changes
- Add jacoco coverage report

# Additional Info
- Sonarcloud coverage checks for a jacoco coverage report therefore we can't use the clover report for sonarcoud.
- Need to specify java file path starting from root for sonar exclusions.
- Need to specify class file path starting from target/class for jacoco exclusions.

- Changed the default branch on Sonarcloud to dev, as we want to compare new code with dev instead of master when we do pull requests.

# Reviewer(s)
 - @Alessandro100 @razine-bensari

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -161,11 +161,6 @@
                 <version>3.7.1</version>
             </plugin>
             <plugin>
-                <groupId>org.openclover</groupId>
-                <artifactId>clover-maven-plugin</artifactId>
-                <version>4.4.1</version>     
-            </plugin>
-            <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
                 <version>0.8.6</version>
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -20,9 +20,10 @@
         <sonar.organization>money-tree</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
         <sonar.coverage.exclusions>
-            src/main/java/com/capstone/moneytree/model/**/*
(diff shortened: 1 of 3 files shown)


wltrf:
The sonarcloud bot says 0% coverage, but that is because it only looks at new code, and the pom.xml is a configuration file and therefore no coverage. This [link](https://sonarcloud.io/dashboard?id=money-tree_BACKEND&pullRequest=152) shows the coverage after merge is 65.8%.

backend/pom.xml
```diff
@@ -164,6 +165,31 @@
                 <artifactId>clover-maven-plugin</artifactId>
```
razine-bensari, line 165: I think we can remove openclover since we are using jacoco in sonarcloud and do not need openclover to generate test coverage reports. Minor detail, it can be removed in another PR too

razine-bensari (approved):
Nice works! thks!!!!

backend/pom.xml
DPigeon: I think we should remove it as it falls under refactoring.

DPigeon:
Nice work!

Issue #135, opened 2020-11-09
Create endpoints for create account


Alessandro100:
As a developer I want to create backend endpoints to handle creating an account

**Acceptance Criteria**
- [x] When given data (email, username, password, first name, last name) it will create a user node in our database
- [x] Must return error if email / username exists exists
- [x] When the user gets an Alpaca key, update the users information in the database **
- [x] Unit tests are written
- [x] The passwords must be encrypted in our database

razine-bensari:
![ssd_flow_createUser](https://user-images.githubusercontent.com/34013446/98755161-0754da00-2396-11eb-99f4-74c20d4af8de.png)

razine-bensari:
Here is a quick draft SSD that the FRONT END will need to conform in order process the user creation

DPigeon:
After starting implementing this story, I realized that #122 was needed before. If we don't want to write ugly and messy code, we need #122 first.

Alessandro100:
If 'sendEmailConfirmation' means that the user will get an email and will have to confirm his account, then we wont be doing that. We are building an MVP

PR #153, opened 2020-11-13
Issue/#135:  Endpoints for Creating an Account

DPigeon:
# Related Issue
- #135 

# Proposed changes
- Endpoint for registering an account
- Endpoint for registering an Alpaca key
- A MoneyTreeSecureString class for encryption/decryption
- Multiple tests that covers most of the paths

# Additional Info
- A user can register an account if he does a POST request on endpoint /api/v1/create-user specifying a User request body object.
- Password is encrypted before storing the user
- A user can register an Alpaca API key if he does a POST request on endpoint /{id}/{key}/register-alpaca-key where id and key are the user id and alpaca key
- Tests for registering an account, registering a key, decryption and encryption

**Note: for better security, we need #141 done so that I can add a secret for the encryption password.**

# Checklist (if applicable)
- [x] Tests

# Reviewer(s)
 - @razine-bensari 
 - @abdulmansour 
 - @wltrf 
 - @Alessandro100


backend/src/main/java/com/capstone/moneytree/controller/UserController.java
```diff
@@ -32,9 +37,59 @@ public UserController(UserService userService) {
...
+
+    HttpStatus validateUserCreation(User user) {
+        String email = user.getEmail();
+        String username = user.getUsername();
+        String password = user.getPassword();
+        String firstName = user.getFirstName();
+        String lastName = user.getLastName();
+
+        // TODO: Refactor this into a validation class validateString method to validate strings & message class too
+        if (email.isEmpty() || email.isBlank() ||
```
razine-bensari, line 82: email needs to be unique, we need to add that in the validation!

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -28,16 +33,87 @@ public DefaultUserService(UserDao userDao) {
...
+    /**
+     * Method to look if user exists with unique email and username.
+     * @param email The email coming from the frontend.
+     * @param username The username coming from the frontend.
+     * @return boolean if exists or not.
+     */
+    @Override
+    public boolean userExists(String email, String username) {
+        boolean exists = false;
+        List<User> userList = userDao.findAll();
```
razine-bensari, line 82: We could do `userDao.findUserByEmail();` and if it returns something then user exists

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -28,16 +33,87 @@ public DefaultUserService(UserDao userDao) {
...
 
-        LOG.info("Created user: {}", createdUser.getFirstName());
+        for (User user : userList) {
+            String dataEmail = user.getEmail();
+            String dataUsername = user.getUsername();
+            if (dataEmail.equalsIgnoreCase(email) || dataUsername.equalsIgnoreCase(username)) {
```
razine-bensari, line 87: This could be moved to the UserValidator class if/when created.

backend/src/test/groovy/com/capstone/moneytree/controller/AlpacaControllerIntegrationTest.groovy
```diff
@@ -31,7 +31,7 @@ class AlpacaControllerIntegrationTest extends Specification {
     def alpacaController
 
     @Shared
```
razine-bensari, line 33: You dont need the @Shared annotation if you are not using that field in a static method (I think lol)

backend/src/test/groovy/com/capstone/moneytree/controller/UserControllerTest.groovy
```diff
@@ -0,0 +1,202 @@
...
+
+@SpringBootTest
+class UserControllerTest extends Specification {
+
+    private UserController userController
+    private DefaultUserService defaultUserService
+    private UserDao userDaoMock
+
+    def setup() {
+        userDaoMock = Stub(UserDao.class)
```
razine-bensari, line 26: You can also use userDaoMock = Mock()

razine-bensari: Just for extra info

razine-bensari (changes requested):
Super Nice work, most of the comments are not applicable if we use a UserValidator class.

Also I offered some suggestion (it does not mean they are better, simply different, let me know what you think).

Some refactoring activity that we can use too.

Amazing work with the tests!!!

backend/pom.xml
```diff
@@ -90,6 +96,12 @@
             <type>pom</type>
             <scope>test</scope>
         </dependency>
+        <dependency> <!-- enables mocking of classes with Spock (in addition to interfaces) -->
+            <groupId>net.bytebuddy</groupId>
```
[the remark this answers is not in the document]
razine-bensari: Actually I dont know what this library does, could you explain it to me ahhaa sorry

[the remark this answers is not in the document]
DPigeon: This library is used to test stubbing classes without using Spock. I have found a way to do the exact same thing with Spock now. I'll remove this dependency because I forgot to.

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
```diff
@@ -32,9 +37,59 @@ public UserController(UserService userService) {
         return users;
     }
 
-    @PostMapping("/users")
-    User createUser(@RequestBody User user) {
-        return userService.createUser(user);
+    @PostMapping("/create-user")
+    ResponseEntity<User> createUser(@RequestBody User user) {
+        ResponseEntity<User> response = null;
+        HttpStatus status = validateUserCreation(user);
```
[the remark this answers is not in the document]
DPigeon: I have a TODO comment about having a Validation story. What we need is a validator class that will validate all possible types. I don't think it falls under this story as validation as a whole would take time. So I think we should create a story for handling validation and use the validation with entities like you talked before cause this is more than just refactoring. What do you think?

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
```diff
@@ -32,9 +37,59 @@ public UserController(UserService userService) {
...
+        return response;
+    }
+
+    /**
+     * A method that updates a user with a new Alpaca key. This should be done only once at beginning
+     * @param userWithKey User with the Alpaca key
+     * @return The new updated user from the database
+     */
+    @PostMapping("/register-alpaca-key")
+    @ModelAttribute
```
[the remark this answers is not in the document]
DPigeon: I forgot to remove that annotation. It is not needed as I was testing something. Thanks 👍

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
```diff
@@ -32,9 +37,59 @@ public UserController(UserService userService) {
...
+    ResponseEntity<User> registerAlpacaApiKey(@RequestBody User userWithKey) {
+        String key = userWithKey.getAlpacaApiKey();
+        if (key.isEmpty() || key.isBlank()) {
```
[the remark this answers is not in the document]
DPigeon: Good point!

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
DPigeon: I don't think we need to add this validation as we assume here that the user is already registered (meaning that the email is already unique). That's why I did not add that validation.

backend/src/test/groovy/com/capstone/moneytree/controller/UserControllerTest.groovy
DPigeon: Yes. Changing!

backend/src/test/groovy/com/capstone/moneytree/controller/AlpacaControllerIntegrationTest.groovy
DPigeon: Yeah we need it otherwise this error comes up:
```
Groovyc: Only @Shared and static fields may be accessed from here
```

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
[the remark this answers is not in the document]
razine-bensari: Yeah you are right, validation should be made in another story, Ill actually create the ticket and add it to the sprint if @Alessandro100  allows it.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
DPigeon: Indeed! Again I think this should be a separate story.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
DPigeon: True. This is even better performance. Thanks!

backend/src/test/groovy/com/capstone/moneytree/controller/AlpacaControllerIntegrationTest.groovy
razine-bensari: Ok ok. I see, thks I did not know

DPigeon:
> Super Nice work, most of the comments are not applicable if we use a UserValidator class.
> 
> Also I offered some suggestion (it does not mean they are better, simply different, let me know what you think).
> 
> Some refactoring activity that we can use too.
> 
> Amazing work with the tests!!!

Thanks! I will update with the changes. Nice suggestions!

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
razine-bensari: but we use that method validateUserCreation() in the createUser() endpoint? I have seen that you still make a email validation in the service though so we do not need to add it here anyway. In anycases, Ill pick up the validation user story and refactor so we have a common place for validation.And from there, people can add more validation classes for the new entities what do you think?

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
razine-bensari: Yeah you are right, definitely too big for this scope

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
DPigeon: Yes I agree!

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
```diff
@@ -56,10 +62,11 @@ public UserController(UserService userService) {
      * @return The new updated user from the database
      */
     @PostMapping("/register-alpaca-key")
-    @ModelAttribute
     public ResponseEntity<User> registerAlpacaApiKey(@RequestBody User userWithKey) {
         String key = userWithKey.getAlpacaApiKey();
-        if (key.isEmpty() || key.isBlank()) {
+
+        // TODO: Refactor this into a validation class validateString method to validate strings & message class too
+        if (StringUtils.isEmpty(key) || StringUtils.isBlank(key)) {
```
razine-bensari, line 69: When using String utils, this: `if (StringUtils.isEmpty(key) || StringUtils.isBlank(key))` becomes this:
`if (StringUtils.isBlank(key))` within that check, both are validated by the library ("" or null). So you dont need to check for both and test coverage will be bigger. It is really minor, no need to update the PR for that.

DPigeon: Done! I did the changes cause we need to improve our test coverage.

backend/pom.xml
```diff
@@ -59,6 +59,12 @@
             <artifactId>spring-boot-autoconfigure</artifactId>
             <version>2.3.4.RELEASE</version>
         </dependency>
+        <!-- Encryption & Decryption -->
+        <dependency>
+            <groupId>com.github.ulisesbocchio</groupId>
+            <artifactId>jasypt-spring-boot-starter</artifactId>
+            <version>3.0.3</version>
+        </dependency>
```
wltrf, line 67: Out of curiosity, why jasypt over spring security?

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -28,16 +33,81 @@ public DefaultUserService(UserDao userDao) {
...
+     * @param username The username of the user
+     * @return The User in the database
+     */
+    @Override
+    public User getUserByEmailAndUsername(String email, String username) {
+        List<User> userList = userDao.findAll();
+        User userToFind = null;
+
+        for (User user : userList) {
+            if (email.equalsIgnoreCase(user.getEmail()) && username.equalsIgnoreCase(user.getUsername())) {
```
wltrf, line 52: Shouldn't this be an OR since we want both unique emails and unique usernames. In this case, if only the email matches, it returns null and vice-versa.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -28,16 +33,81 @@ public DefaultUserService(UserDao userDao) {
...
+    }
 
-        LOG.info("Created user: {}", createdUser.getFirstName());
+    public String encryptData(String text) {
+        return secureString.toGraphProperty(text);
+    }
 
-        return createdUser;
+    public String decryptData(String text) {
+        return secureString.toEntityAttribute(text);
```
wltrf, line 111: Do we need this? We shouldn't have to decrypt password and manipulate the unencrypted string, it is bad practice.

backend/src/main/java/com/capstone/moneytree/utils/MoneyTreeSecureString.java
```diff
@@ -0,0 +1,43 @@
...
+        return stringEncryptor().encrypt(text);
+    }
+
+    /**
+     * Method to decrypt the value before it is used in the backend
+     * @param text The string value
+     * @return A decrypted text
+     */
+    public String toEntityAttribute(String text) {
+        return stringEncryptor().decrypt(text);
```
wltrf, line 41: similar to previous comment, we don't need this unless I'm missing something.

wltrf (commented):
I left some comments, razine made some good points as well.

wltrf:
Also, we should be using a password encryptor over a string encrypter as the string encryptor doesn't provide utility to compare a string with an encrypted string. See jasypt docs [here](http://www.jasypt.org/encrypting-passwords.html)

razine-bensari:
> Also, we should be using a password encryptor over a string encrypter as the string encryptor doesn't provide utility to compare a string with an encrypted string. See jasypt docs [here](http://www.jasypt.org/encrypting-passwords.html)

Yeah you are right, but to be honest, I have never used password encryptor or string encryptor. Do you think we should address that in this ticket or create another ticket to address those changes? (Refactoring maybe?) @wltrf

backend/pom.xml
DPigeon: I used jasypt because of this article https://www.geeksforgeeks.org/how-to-encrypt-passwords-in-a-spring-boot-project-using-jasypt/.

backend/src/main/java/com/capstone/moneytree/utils/MoneyTreeSecureString.java
DPigeon: I wanted to compare password for the unit test I made. However I realized that this is not how we compare encrypted passwords. And also we won`t really decrypt a password if needed. I`ll remove this.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
DPigeon: True. I thought initially that when we login in the future we would compare the decrypted passwords but it is not recommended to do that and we should compare the digests instead.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
DPigeon: Right! I'll make the changes.

DPigeon:
> Also, we should be using a password encryptor over a string encrypter as the string encryptor doesn't provide utility to compare a string with an encrypted string. See jasypt docs [here](http://www.jasypt.org/encrypting-passwords.html)

You are right. I thought it was doable with a StringEncryptor. I'll make the appropriate changes.

razine-bensari (approved):

wltrf (approved):
Nice, thanks for the changes.

DPigeon:
I have **refactored** the alpaca register key endpoint to use a POST request specifying a user ID with a key instead of using a User object with a key in it.

Issue #137, opened 2020-11-09
Add elements to the login page


Alessandro100:
As a user I would like to see all the elements of the frontend page to better navigate

**Acceptance Criteria**
- The frontend page must look like the mockups
- Must be mobile accessible

![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/aae85f43-f6ec-48bc-9882-248b1df3ca53)

PR #162, opened 2020-11-20
Feature#137/add elements to the login page

arthur-del:
#137  https://app.zenhub.com/workspaces/moneytree-board-5f5bcf1c7ea48f001c28c973/issues/dpigeon/money-tree/137

# Proposed changes
- desktop version scales down to mobile when 600px threshold reached
- Kept the image background the same for desktop and mobile for consistency sake

# Additional Info
- There's a bug that appears if you scale the mobile screen to "iphone 6/7/8 plus" a small line appears over the input boxes,
any insight is welcomed! I'll be creating a bug report once this PR is merged into dev

# Checklist (if applicable)
- [x ] Tests
- [ ] Documentation

# Reviewer(s)
 - @Alessandro100 
 - @EspressoCode 
 - @marwan011 
Any requested changes will most likely be addressed Monday morning :^)


client/src/app/components/signup-form/signup-form.component.scss
```diff
@@ -0,0 +1,34 @@
...
+    padding: 16px;
+
+    .signup-title {
+      text-align: center;
+      margin-bottom: 16px;
+      font-size: 30px;
+      color: $mt-white;
+    }
+  }
+}
```
EspressoCode, line 34: adding these stylings will override the default styling of mat-form-field outline, which can be a solution for removing the extra blue line in the form. 
```
::ng-deep .mat-form-field-appearance-outline .mat-form-field-outline {
      background-color: $mt-white;
      border-radius: 5px;
    }
```
I myself had many problems with the styling of mat-input elements, I looked on StackOverflow and other websites to find solutions and I saw that many people had different problems using it, the default styling is nice but I found it very hard to customize!
I also see that if we use the autocomplete feature of chrome (when you already saved your name and put it in the filed by Chrome), it shows a separate, smaller, weird box inside the input box, so maybe we should use regular input box in the end if we had so many problems with mat-input.

client/src/app/pages/login-signup/login-signup.component.scss
```diff
@@ -0,0 +1,98 @@
+@import "../../../styles.scss";
+
+.container {
+    mix-blend-mode: screen;
```
EspressoCode, line 4: I removed this line and tried the home page and I couldn't see a change, so maybe it is safe to delete this line?

client/src/app/app-routing.module.ts
```diff
@@ -2,10 +2,12 @@ import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
 import { StockDetailComponent } from './pages/stock-detail/stock-detail.component';
 import { LeaderboardComponent } from './pages/leaderboard/leaderboard.component';
+import { LoginSignupComponent } from '../app/pages/login-signup/login-signup.component';
 
 const routes: Routes = [
   { path: 'stock-detail/:ticker', component: StockDetailComponent },
   { path: 'leaderboard', component: LeaderboardComponent },
+  { path: 'home', component: LoginSignupComponent} // TO DO: choose adequate path for landing page
```
EspressoCode, line 10: I discussed this point with @Alessandro100, if we choose home for our sign-in/sign-up page, it would confuse us in future, as we have a home button in our header that takes the user to the first page of the application (the one that has all the news-feed and right and left side panels, the page that user sees after signing in). If we change it now, it would be easier, I think maybe we can change the route name for this page to sing-up or welcome page.

EspressoCode (commented):
Great work 👍🏻 , I added a few comments.

client/src/app/app-routing.module.ts
arthur-del: In that case would an empty route be the better option? 
So that the default is going to this page but if the user is already logged in we redirect to his home actual home page. 
I do not have much knowledge about internet cache so correct me if this wrong
@Alessandro100

client/src/app/pages/login-signup/login-signup.component.scss
arthur-del: I think you're right, I had improper implementation of that css styling

client/src/app/components/signup-form/signup-form.component.scss
arthur-del: @Alessandro100

marwan011: we can override the styling for all mat-input without using ng:deep ... i ll make the change as soon as possible

Alessandro100: ok so @EspressoCode I looked at the code and the lines you highlighted are not related to Angular Material components. In terms of overriding the style for the mat-input, I read the docs and it seems like the best way to style this is with ng-deep. @marwan011 if you find a more elegant solution, we can go over it together, but we want the finish that angular material brings (animation when clicking, color when focus, etc)

client/src/app/app-routing.module.ts
Alessandro100: you are right, we will change the route to empty

Alessandro100 (approved):
Great stuff Arthur! The page really looks like the figma document, well done

Issue #155, opened 2020-11-15
Create user validator


razine-bensari:
In order for Spring to handle validation in a decoupled manner and to focus to core business logic within our controller, we can use the @Valid annotation class and validate our entities that we receive from front end.

The goal of this ticket is to create an User validator class and make that classe's responsability to validate the user object using spring's validator feature.

PR #163, opened 2020-11-22
Issue/#155 user validator

razine-bensari:
# Related Issue
- #155 

# Proposed changes
- addition of validator interface to implement the entity validators
# Additional Info
- any additional information or context

# Checklist (if applicable)
- [CHECK] Tests

# Reviewer(s)
 - @DPigeon @wltrf @abdulmansour


backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -1,126 +1,124 @@
...
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
 import javax.security.auth.login.CredentialNotFoundException;
-import java.util.List;
 
+/**
+ * {@inheritDoc}
```
DPigeon, line 21: Is this line for the API doc to do from another story?

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
```diff
@@ -41,75 +50,56 @@ public UserController(UserService userService) {
...
+    * A POST method that receives a user JSON object and registers it
+    *
+    * @param user The JSON object body
+    * @return A proper response with the URI of the newly created user
+    */
+   @PostMapping("/create-user")
+   public ResponseEntity<User> createUser(@RequestBody User user) {
+      User createdUser = userService.createUser(user);
+
+      URI userURI = ServletUriComponentsBuilder.fromCurrentContextPath().
```
DPigeon, line 63: Can you explain what this line does?

backend/src/main/java/com/capstone/moneytree/validator/UserValidator.java
```diff
@@ -0,0 +1,80 @@
...
+
+   private final UserDao userDao;
+
+   @Autowired
+   public UserValidator(UserDao userDao) {
+      this.userDao = userDao;
+   }
+
+   @Override
+   public boolean supports(Class<?> aClass) {
```
DPigeon, line 28: Just out of curiousity, when would this validation be useful?

DPigeon (commented):
Nice work! I see how the flow for validation is and it is pretty neat. So every service can be validated with its own class. Thanks for the test utils class!

I left some minor comments. 
Also, I think we should have unit tests for the UserValidator class. We're mocking it in the UserControllerTest class, but we should have test it with a real instance too with unit tests. What do you think?

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
razine-bensari: So based on HTTP protocol, when a resource is created, we should return 201 Created with a header that contains the url of the resource. Some client will not support receiving 200 when they create a resource. It is just a matter of following the protocol

backend/src/main/java/com/capstone/moneytree/validator/UserValidator.java
wltrf: For now I think probably not much use for us, but it has to be overriden since its part of the Validator interface. I searched a bit and found this [doc link](https://docs.spring.io/spring-framework/docs/4.1.7.RELEASE/spring-framework-reference/html/validation.html) where it explains bean validation. There's a nice example where they check in the constructor if the validator class supports the object passed, which is good to return more meaningful error messages. The support can be implemented to either return true for the class itself, or tweaked to allow reuse with other classes and/or inheritance stuff. Good comment.

backend/src/main/java/com/capstone/moneytree/validator/Validator.java
```diff
@@ -0,0 +1,15 @@
...
+ * classes for our entities.
+ */
+public interface Validator {
+
+   boolean supports(Class<?> var1);
+
+   void validate(Object var1);
+
+   boolean isValid(Object var1);
+}
```
wltrf, line 15: We could use the [spring validator interface](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/validation/Validator.html)

wltrf (approved):

backend/src/main/java/com/capstone/moneytree/validator/Validator.java
razine-bensari: Thks for the link, I actually tried to do so before implementing our own interface which looks super simialr to the spring one. The reason why I settle by providing our own interface is that in the spring interface, within the validate method, we need to provide an error object. But the way we design our rest api is that we do no deal with errors, we deal with exception and have our exception handlers return the response to the user.

That is why we have our own interface, only difference is that the validate() methods takes in an object to validate with no error object as opposed to spring.

backend/src/main/java/com/capstone/moneytree/validator/UserValidator.java
razine-bensari: https://github.com/DPigeon/Money-Tree/pull/163/commits/6cb6a7711029b596897c6465830a57f2f9931112

razine-bensari:
> Nice work! I see how the flow for validation is and it is pretty neat. So every service can be validated with its own class. Thanks for the test utils class!
> 
> I left some minor comments.
> Also, I think we should have unit tests for the UserValidator class. We're mocking it in the UserControllerTest class, but we should have test it with a real instance too with unit tests. What do you think?

https://github.com/DPigeon/Money-Tree/pull/163/commits/6cb6a7711029b596897c6465830a57f2f9931112

Replied to comment and updated PR commits

Issue #168, opened 2020-11-23
CORS Not Working Properly on UAT / PROD environment


DPigeon:
**Describe the bug**
CORS is not working properly.

**To Reproduce**
1. Go to dev.money-tree.tech
2. Try this endpoint http://dev.money-tree.tech/stock-detail/APPL
3. Error in the dev options

**Expected behavior**
We should be able to make calls from the frontend to the backend and see the data fetched.

**Screenshots**
![bug](https://user-images.githubusercontent.com/37888675/100008658-b84a7400-2d9b-11eb-9b3b-1c732d514f9d.png)

**Desktop (please complete the following information):**
 - OS: Windows

Bug reported by @DPigeon

DPigeon:
@razine-bensari Hossein and I tried @CrossOrigin("localhost:4200") locally and cors plugin was not needed anymore when fetching from the frontend. We may need to use something different than localhost when deploying like @CrossOrigin("127.0.0.1").

DPigeon:
This bug is still not fixed. I will open it again.

PR #170, opened 2020-11-23
Hotfix for cors

razine-bensari:
# Related Issue
- #168 

# Proposed changes
- changing dns name

# Reviewer(s)
 - @DPigeon

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/controller/MoneyTreeController.java
+++ b/backend/src/main/java/com/capstone/moneytree/controller/MoneyTreeController.java
@@ -13,6 +13,6 @@ import org.springframework.web.bind.annotation.RestController;
 @Retention(RetentionPolicy.RUNTIME)
 @RestController
 @RequestMapping("/api/v1")
-@CrossOrigin(value = "http://localhost")
+@CrossOrigin(origins = "http://127.0.0.1")
 public @interface MoneyTreeController {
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/capstone/moneytree/controller/MoneyTreeController.java
```diff
@@ -13,6 +13,6 @@
 @Retention(RetentionPolicy.RUNTIME)
 @RestController
 @RequestMapping("/api/v1")
-@CrossOrigin(value = "http://localhost")
+@CrossOrigin(origins = "http://127.0.0.1")
```
DPigeon, line 16: We also need to be able to test with CORS locally. Add this other origins too like follows:
@CrossOrigin(origins = {"http://localhost:4200", "http://127.0.0.1"})

wltrf (approved):
If this becomes too much of a problem, aka we keep encountering bugs related to that, we could ditch checking cors in the app and do it only when we deploy with docker.

backend/src/main/java/com/capstone/moneytree/controller/MoneyTreeController.java
razine-bensari: nice catch

DPigeon (approved):

Issue #154, opened 2020-11-13
CI Workflow Improvements


wltrf:
The current ci.yml workflow is slow and could be improved.

1. Split ci.yml into 2 files: ci_frontend.yml and ci_backend.yml
This would allow us skip one of the ci if there are no changes in that particular folder.
The way this works is by specifying a path for the action trigger.
e.g. for ci_backend.yml
```yml
on:
    pull-request:
        paths:
        -'./backend/**'
```
see this [link](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) for more information.

2. We currently run the same maven goals multiple times. The way maven works is different than npm in that maven will call all goals from the lifecycle up and until the specified goal while npm just runs the specific script.
See this [link](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html)
https://github.com/DPigeon/Money-Tree/blob/df28d283dd6109af2e28427bce1aed34df5d34d1/.github/workflows/ci.yml#L96-L107
The code above will run the following goals:
- validate
- compile
- test
- package
- verify
- **install**
- **clean**
- validate
- compile
- **test**
- validate
- compile
- test
- package
- **verify**

This is clearly redundant and could be replaced by a single step which calls `mvn clean verify`

3. This isn't a performance issue, but more of a simplification question: Why do we use the ./mvnw instead of using the mvn binary provided by github actions i.e. use mvn instead of ./mvnw

** ACCEPTANCE CRITERIA **
- pipeline split into two files, see 1.
- pipeline runs only if a change in the specific folder, see 1.
- there is a single maven call for backend, see 2.
- there is no instance of mvnw (wrapper) only the mvn (binary), see 3.

wltrf:
@DPigeon thoughts, comments?
@razine-bensari am i wrong on the maven stuff?

DPigeon:
Yes everything inside this is good! We can improve the pipeline for sure.
1. Totally agree
2. Redundant for sure. However how are we going to validate if the cache exists with a single mvn clean verify command?
3. The mvn command was not working in the cicd environment when I first worked on it.

wltrf:
Cache is done automatically in the step before. We are currently the non wrapper mvn for the sonarscanner step, so it works.

PR #184, opened 2020-11-27
#154 ci improvements

wltrf:
# Related Issue
- #154 

# Proposed changes
- Split ci.yml into 2 files: ci_frontend.yml and ci_backend.yml
- Single step which calls `mvn clean verify` for backend instead of many redundant repetitive and slow
- use mvn (binary) instead of ./mvnw (wrapper)
- run ci on push to dev and master so we get sonarcloud scan for those branches without having to do manually
- some minor cache handling redundencies and extra unecessary npm install step removed
- use npm ci instead of npm i see [here](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable)

# Additional Info
- ci_frontend will run when any change in client/**
- ci_backend will run when any change in backend/**

# Reviewer(s)
 - @DPigeon 
 - @Alessandro100


wltrf:
here we can see that a change in the /backend folder triggered the CI Backend Pipeline only and a single sonarcoud bot comment was registered.
Screenshot from 2020-11-27 14-13-37

wltrf:
here we see that changes in both frontend and backend will result in both ci pipelines to trigger with two comments from the sonarcloud both just as before.
Screenshot from 2020-11-27 14-22-30

wltrf:
similarly to backend, if theres a change only on frontend, only the frontend ci pipeline will run
Screenshot from 2020-11-27 14-33-32

wltrf:
Now, the only issue is when we don't have any changes in either ./backend or ./client, the checks will wait and never be fulfilled (note the lack of a check mark/yellow circle/red circle next to the verified label). A workaround is to modify the readme in backend or client with a minor change that doesn't have any impact like a space.
Screenshot from 2020-11-27 14-51-12

wltrf:
This last commit reflect the workaround explained in the previous comment. (see the yellow circle now ;) )
Screenshot from 2020-11-27 14-56-04

Alessandro100 (approved):
Looks good to me, nice changes. One small note is when making a branch name just format it to our convention
issue/...
feature/....
bug/...

other than that it's good to go

Issue #178, opened 2020-11-27
Investigate CORS/ActuatorHealth Issues on the UAT / PROD environment


Alessandro100:
As a developer I would not like to have any CORS issues in the UAT / PROD environment so I can see actual data

**Acceptance Criteria**
- There are no CORS bugs in the UAT / PROD env

DPigeon:
@razine-bensari I investigated a bit on this and it seems like CORS is working fine now (we don't get any CORS error if we go to http://dev.money-tree.tech/stock-detail/ac and look at the browser console in dev options). Indeed, we're getting a CONNECTION_REFUSED because we did not setup any Neo4j database on any environments in the VMs except for the testing database. 

So we'll have to do that.

DPigeon:
One solution I see after reading [this](https://docs.docker.com/compose/networking/): 
- Use docker compose on the VMs with backend and neo4j linked together
- Inject the production profile when deploying on uat/prod
- Then in the prod properties we can use IMAGE_NAME://SERVICE_NAME:7687 in the prod properties like
```
spring.data.neo4j.uri=neo4j://neo4jdb:7687 
```
- Now our backend container will be able to talk to our neo4j container correctly since they will be on the same network using the proper container IP (neo4jdb refers to the docker IP of the service container). 

The reason we should use docker compose is that without it, we'll have to put the specific docker IP of the neo4j container that is on the same network. That would be something like 172.10.0.2 for example (starting with 172). The properties would become
 ```
spring.data.neo4j.uri=bolt://172.10.0.2:7687
```
And that can become messy especially if we use uat and prod with the same prod-property file because both neo4j containers on the uat and prod VMs may resolve to different IPs. Using docker compose helps us "hide" the IP of the container and call it with its service name instead.

I am still investigating. I know this is far from being a CORS problem so we may change the title of this issue accordingly.

razine-bensari:
I see, thks for the research, indeed, using docker compose will create an internal network for all services within the docker-compose file. I guess we can do that within that ticket

PR #187, opened 2020-12-03
Bug: #178 Investigate CORS & Actuator Issue

DPigeon:
# Related Issue
- #178 

# Proposed changes
- Neo4j Database linked on the same network with our backend (fixes the actuator)
- Client linked on the same network with our backend (fixes CORS & bad gateway calls)
- CD changes related to the angular being on uat or prod
- nginx reverse proxy added and now all backend calls should be made with the http://ENV_URL/api/ as a baseUrl. @Alessandro100 I'll let you do this part on your end. I left a comment for you to do it. 
- Deploy.sh scripts updated on the VMs ready to start docker-compose. Here is a small snippet of what was added. $3 is the version name obtained from the CD to start the proper docker image with its tag.
```
export APP_VERSION="$3"
APP_VERSION=$3 docker-compose up -d --force-recreate
```

# Additional Info
So here is the idea (along with this proposed solution [here](https://github.com/DPigeon/Money-Tree/issues/178#issuecomment-737359300)):
Docker-compose was used here to configure our frontend, backend and database on the same network. You can read more on it [here](https://docs.docker.com/compose/networking/). Basically, using docker-compose allows us to use services. These services encapsulate the local IP docker set for each container. It is therefore easier to use when it comes to make calls from the frontend to the backend and from the backend to the database in UAT or production environment. For example, instead of calling http://DOCKER_CONTAINER_IP:8080 from the frontend, the container can simply call http://SERVICE_NAME:8080 without having to worry about IP addresses changing.

# Reviewer(s)
 - @razine-bensari 
 - @wltrf 
 - @Alessandro100

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -17,4 +17,4 @@ RUN mkdir jar
 # Copy jar file generated 
 COPY --from=maven /home/maven/target/moneytree-*.jar /home/maven/jar/moneytree.jar
 # Execute command once image started
-CMD ["java", "-jar", "/home/maven/jar/moneytree.jar"]
+CMD ["java", "-Dspring.profiles.active=prod", "-jar", "/home/maven/jar/moneytree.jar"]
--- a/backend/src/main/resources/application-prod.properties
+++ b/backend/src/main/resources/application-prod.properties
@@ -11,7 +11,7 @@ spring.data.neo4j.password=${NEO4J_PASSWORD}
 
 # Alpaca Related properties
 # Client ID Auth: 198903d0d2f523a25e4dce65837bbf0d
-# Client Secret Auth: d756cb7ae52ad380d16aec004a38b00f04747cf2docker
+# Client Secret Auth: d756cb7ae52ad380d16aec004a38b00f04747cf2
 alpaca.api.version=v2
 alpaca.key.id=${ALPACA_KEY_ID}
 alpaca.secret=${ALPACA_SECRET_ID}
--- a/backend/src/main/resources/application-prod.properties
+++ b/backend/src/main/resources/application-prod.properties
@@ -5,13 +5,13 @@ server.servlet.context-path = /api/v1
 
 server.port=8080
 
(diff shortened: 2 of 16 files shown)


backend/src/main/resources/application-prod.properties
```diff
@@ -5,7 +5,7 @@ server.servlet.context-path = /api/v1
 
 server.port=8080
 
-spring.data.neo4j.uri=bolt://localhost:7687
+spring.data.neo4j.uri=neo4j://neo4j:7687
```
wltrf, line 8: just a comment, neo4j:// instead of bolt:// simply allows to have a neo4j cluster instead of only a single instance. We won't make use of that, but its good. the neo4j: part is from the docker-compose file (there's a link: part) You already knew this but this is a comment for others if they are wondering.

backend/Dockerfile
```diff
@@ -17,4 +17,4 @@ RUN mkdir jar
 # Copy jar file generated 
 COPY --from=maven /home/maven/target/moneytree-*.jar /home/maven/jar/moneytree.jar
 # Execute command once image started
-CMD ["java", "-jar", "/home/maven/jar/moneytree.jar"]
\ No newline at end of file
+CMD ["java", "-Dspring.profiles.active=prod", "-jar", "/home/maven/jar/moneytree.jar"]
```
wltrf, line 20: could use env variable instead of hardcoding, still works like that.

wltrf (approved):
good work.

backend/src/main/resources/application-prod.properties
DPigeon: Yes indeed! The link is provided in the dockerfile. So the backend service is linked with the neo4j database service.

backend/Dockerfile
DPigeon: I made it like that because we only have 2 profiles in the backend (dev and prod). If one day we expand on those, we can make the proper changes. That dockerfile is used to deploy so the profile will always be prod for now.

Issue #188, opened 2020-12-04
Fix the IEXCloud Tests


DPigeon:
**Describe the bug**
Two IEXCloud tests are failing on the backend.

**To Reproduce**
1. Remove the @ Ignore for tests in StockControllerIntegrationTest.groovy.
2. Run mvn (./mvnw) test locally or push a commit without the Ignore annotation in StockControllerIntegrationTest
3. See the error in the console

**Expected behavior**
Both tests should pass. This should be investigated to see what's happening. The Ignore annotation was added to both tests for now.

**Screenshots**
![1](https://user-images.githubusercontent.com/37888675/101962696-d35e1600-3bda-11eb-9d1d-ca52b1004f9d.png)

**Desktop (please complete the following information):**
 - Windows

Bug reported by @DPigeon

DPigeon:
This can be closed as it has been fixed before.

PR #189, opened 2020-12-04
#190: Hotfix/fix iexcloud bug

razine-bensari:
# Related Issue
- Fixes bug #190 

# Proposed changes
- update iexcloud library to change the type of the date field in KeyStats to String instead of LocalDateTime because,
when using IECLoud sandbox environment, the api returns "0" as a date which throws exception since the library tries to parse this field to a date.

 - @DPigeon @marwan011 @EspressoCode

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -130,7 +130,7 @@
         <dependency>
             <groupId>com.github.razine-bensari</groupId>
             <artifactId>iextrading4j</artifactId>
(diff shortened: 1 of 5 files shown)


DPigeon (approved):
Approved! And the tests will be fixed in another story #188 as this needed to be fixed asap. It was blocking other stories on the frontend.

marwan011 (approved):
Well done @razine-bensari

Issue #199, opened 2020-12-09
Nginx Proxy Pass Forbidden Access 403


DPigeon:
**Describe the bug**
Proxy pass on Nginx gives forbidden access 403.

**To Reproduce**
1. Go to http://dev.money-tree.tech/
2. Try to do any POST request (like signup)
3. See error 403 in console log

**Expected behavior**
We should be able to send requests from the client to proxy pass backend. For some reasons, the current proxy does not allow it.

**Screenshots**
![Untitled](https://user-images.githubusercontent.com/37888675/101655527-b6c1b280-3a0f-11eb-85c7-33c860ae1e0e.png)

**Desktop (please complete the following information):**
 - Windows Google Chrome

Bug reported by @DPigeon

DPigeon:
Note: this may require multiple uat tests unfortunately. It cannot be tested locally as we need the baseUrl to be dev.money-tree.tech. It may also be a permission error on the VM's side.

DPigeon:
After testing on http://dev.money-tree.tech/ when signing up, the user was successfully added to the http://dev.money-tree.tech:7474/browser/ dev db.

DPigeon:
Tested and works!

PR #201, opened 2020-12-10
Bug/#199 Nginx Proxy Pass Forbidden

DPigeon:
# Related Issue
- #199 

# Proposed changes
- Nginx config changed for proxy pass. The proxy should allow request coming from the client.

# Screenshots (if applicable)

# Reviewer(s)
 - @Alessandro100 
 - @razine-bensari 
 - @wltrf

### DIFF
--- a/client/nginx.conf
+++ b/client/nginx.conf
@@ -9,9 +9,8 @@ server {
     }
 
     location /api {	
-        proxy_set_header X-Forwarded-Host $host;	
-        proxy_set_header X-Forwarded-Server $host;	
-        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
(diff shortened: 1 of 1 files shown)


wltrf (changes requested):
Needs further testing on dev to know if this works before merging. @DPigeon you can deploy manually on the vm and test if it works by bypassing our deploy pipeline, it will be more efficient that way than merging multiple times

Alessandro100 (approved):
As this is something I cannot test locally, I will approve it as it looks good, but we will keep an eye on it's functionality

DPigeon:
Yeah I'll keep changing this PR with the VM tests.

wltrf (approved):

Issue #167, opened 2020-11-23
Input field get random blue line 


arthur-del:
**Describe the bug**
A blue vertical line appears on the left hand side of input fields under certain emulated mobile sizes

**To Reproduce**
1. go to base URL
2. open developer tools 
3. select mobile design iphone 6/7/8 plus or Ipad Pro or pixel 2

**Expected behavior**
Input fields should be fully white

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5e1e14c10fe859d66dd1d06d/23f46fb7-a707-485f-9fb8-de37096c276d)

**Smartphone (please complete the following information):**
 - Browser: Chrome

**Additional context**
This problem is related to angular material and how we customize those templates

Bug reported by @arthur-del

PR #217, opened 2020-12-14
fixed the Input field blue line  bug  #167

EspressoCode:
# Related Issue
- This pr is a bug fix for #167 Input field get random blue line.

# Proposed changes
- Fixed the problem of the extra blue line by adding custom styles to .mat-form-field-flex class of mat-form-field.

# Additional Info
- Since the issue is only on the mobile version and specifically for signup form fields (when on a dark background), I put the changes only in styling file of that component and not in forms styling. Should we have problems with other forms we might move the changes to that file in future!

# Reviewer(s)
 - @Alessandro100 @arthur-del

### DIFF
--- a/client/src/app/components/signup-form/signup-form.component.scss
+++ b/client/src/app/components/signup-form/signup-form.component.scss
@@ -35,9 +35,5 @@
       font-size: 26px;
       color: $mt-white;
     }
-    ::ng-deep .mat-form-field-flex{
-      background-color: white;
-      border-radius: 5px;
-    }
   }
 }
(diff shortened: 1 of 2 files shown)


client/src/app/components/signup-form/signup-form.component.scss
```diff
@@ -35,5 +35,9 @@
       font-size: 26px;
       color: $mt-white;
     }
+    ::ng-deep .mat-form-field-flex{
+      background-color: white;
+      border-radius: 5px;
+    }
```
Alessandro100, line 41: Since this is a bug fix that should affect all input boxes and not only the ones in signup-form, this class should be moved to 'themes.scss'. But the fix works great!

Alessandro100 (changes requested):
Just 1 minor comment. The fix works great, just need to refactor so it applies throughout the application

Alessandro100 (approved):
Looks good nice changes

Issue #204, opened 2020-12-10
User is able to click 'signup' multiple times


Alessandro100:
**Describe the bug**
When a user is creating an account, when they click 'signup' they can click it many times and create many instances of themselves

**To Reproduce**
1. Go to the signup page
2. Enter all your information to signup
3. Click the 'signup' btn many times
4. In the database it will create many instances of that user

**Expected behavior**
The user should only be able to click the button once and have it followed by a loading icon

**Screenshots**
Double instance of kalessandro
![image](https://user-images.githubusercontent.com/18631060/101830496-97578200-3b02-11eb-9fe2-4d683bbb2abd.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser chrome

**Additional context**

Bug reported by @Alessandro100

PR #219, opened 2020-12-18
Bug/#204 user is able to click 'signup' multiple times

EspressoCode:
# Related Issue
- #204 user is able to click 'signup' multiple times

# Proposed changes
This pr fixes the problem of multiple clicks on signup and login buttons. Now when the user clicks on sign up or login buttons, the button become unclickable (disabled) and it becomes available only in case that we have an appError which means the signup/login forms are filled out with valid entries but the response from backend shows that authentication/user creation failed.

# Reviewer(s)
 - @Alessandro100


client/src/app/components/signup-form/signup-form.component.html
```diff
@@ -22,14 +22,17 @@ <h1 class="signup-title">Sign Up</h1>
...
 
-        <button id="signup-submit-btn" class="mt-input-field" mat-flat-button type="submit" [class.btn-opacity-disabled]="signUpForm.invalid"
-            [disabled]="!signUpForm.valid">
+        <button id="signup-submit-btn" class="mt-input-field" mat-flat-button type="submit"
+            [class.btn-opacity-disabled]="signUpForm.invalid"
+            [disabled]="(!signUpForm.valid || submitted) && !appError">
             SIGN UP
         </button>
-
+        <div class="form-error-messages" *ngIf="appError">A user with this email address or username already exist!
```
arthur-del, line 34: Should we specify which between the username and the email  already exists? I feel as though signing up should be as smooth as possible of process.

client/src/app/components/login-form/login-form.component.html
```diff
@@ -8,8 +8,9 @@ <h2 class="login-text-hint">Already have an account?</h2>
     <mat-form-field class="mt-input-field half" appearance="outline" color="accent">
       <input id="login-password" matInput type="password" placeholder="Password" [formControl]="pwd" />
     </mat-form-field>
-    <button id="login-submit-btn" mat-flat-button class="mt-input-field half" [class.btn-opacity-disabled]="logInForm.invalid"
-      [disabled]="logInForm.invalid">LOGIN</button>
+    <button id="login-submit-btn" mat-flat-button class="mt-input-field half"
+      [class.btn-opacity-disabled]="logInForm.invalid"
+      [disabled]="(logInForm.invalid || submitted) && !appError">LOGIN</button>
```
Alessandro100, line 13: on line 13, there is a lot of logic going on in the [disabled] box, we can move this to a function and test it.

client/src/app/components/signup-form/signup-form.component.html
```diff
@@ -22,14 +22,17 @@ <h1 class="signup-title">Sign Up</h1>
...
-            <input id="signup-password-2" matInput placeholder="Confirm Password" type="password" [formControl]="pwd2" />
+            <input id="signup-password-2" matInput placeholder="Confirm Password" type="password"
+                [formControl]="pwd2" />
         </mat-form-field>
 
-        <button id="signup-submit-btn" class="mt-input-field" mat-flat-button type="submit" [class.btn-opacity-disabled]="signUpForm.invalid"
-            [disabled]="!signUpForm.valid">
+        <button id="signup-submit-btn" class="mt-input-field" mat-flat-button type="submit"
+            [class.btn-opacity-disabled]="signUpForm.invalid"
+            [disabled]="(!signUpForm.valid || submitted) && !appError">
```
Alessandro100, line 31: Same comment as above, there is a lot of logic going on here that we can test

Alessandro100 (changes requested):
I left some changes regarding refactoring / testing logic. Also I noticed that some of the errors are overlapping, we should look into that
image

Alessandro100:
oh another thing, don't use quotation marks in a branch name! It wont be simple to copy past since we will have to do \'

EspressoCode:
> oh another thing, don't use quotation marks in a branch name! It wont be simple to copy past since we will have to do\ '


ok sorry, I will keep that in mind for next time.

client/src/app/components/signup-form/signup-form.component.html
EspressoCode: that is the right way, I agree! But for now, we have a generic message for both, in future, we will separate the two after getting this one working properly.

EspressoCode:
I used the new code in my PR for error handling for fixing the issue of multiple clicks, please review that PR first, so that if there are changes to that one I can apply them here. I resolved all the issues and moved a part of the logic to functions, I will add tests for new parts soon and test login/sign up thoroughly again this time.

Alessandro100 (approved):
Looks good, nice changes

PR #220, opened 2021-01-09
Issue/#180 add spring profile

razine-bensari:
# Related Issue
- #180 

# Proposed changes
- addition of multiple properties file, configured to be passed via command line when starting the app locally

We can now use those command to run the app locally with the given spring profil:

mvn spring-boot:run -Dspring.profiles.active=prod
mvn spring-boot:run -Dspring.profiles.active=local
mvn spring-boot:run  (uses dev db)

# Checklist (if applicable)
non applicable

TO BE MERGED AFTER PR #218 

# Reviewer(s)
 - @wltrf 
 - @abdulmansour 
 - @DPigeon

Issue #165, opened 2020-11-23
Change Existing Tests to Use Test DB Instead of Mocking.


wltrf:
As a developer I would like to use test db instead of mocking data for better tests

See #161 

**Acceptance Criteria**
- Change the tests to not use mocking stuff, but rather use a test db

wltrf:
#180 related

DPigeon:
Note that for this story, we`ll be using the UAT database for tests [here](http://dev.money-tree.tech:7474/browser/).

wltrf:
^ additional info on comment above, this is because we cap our memory limit on the vm when running 2 db instances. We can use different schemas inside the same dbms instead.

DPigeon:
Yes so for more information, we tried running 2 neo4j instances but since we're already running the backend and client, we peaked 95% memory. See this graph below for the uat VM.
![2222](https://user-images.githubusercontent.com/37888675/101649481-ecaf6880-3a08-11eb-9448-8e86bc928617.png)

The sudden drop is one of the neo4j instances shutting down due to a lack of memory.

wltrf:
**Summary**
1. Can't run 2 dbms due to vm memory limitations
2. Can't create a separate db inside the existing dbms due to it being an "enterprise" feature

Solution: Use the uat database, and make prefix users with test.

wltrf:
Note: We added a 1G swapfile to the vm to alleviate memory and fix some crashes and issues.

DPigeon:
Hey @wltrf and @razine-bensari are the two branches you worked on for this issue the integrated together? It seems like this branch made from Walter https://github.com/DPigeon/Money-Tree/tree/feature/%23165-test-db and this PR made from Razine https://github.com/DPigeon/Money-Tree/pull/221 have no link between them.

razine-bensari:
@DPigeon  @wltrf  was exploring on how to have an internal db running with the test for integration testing. He proposed to simply connect to the UAT db instead as having the internal seems to not work because of emmory  limitations.

The pr I made is to use the UAT db in integration testings

DPigeon:
Ok thanks for clarifying!

PR #221, opened 2021-01-09
Issue/#165 test db for spock

razine-bensari:
# Related Issue
- #165 

# Proposed changes
- added some pom.xml config and a library to instanciate the gorm driver and connect to local db when doing integration tests.

# Additional Info
Unit test files follow this pattern `*Test.groovy`
Integration test file follow this pattern `*IT.groovy`

To run unit test:
`mvn clean test`

To run integration test
`mvn failsafe:integration-test`

When executing `mvn verify` both unit and integration tests will be executed after the packaging phase of the jar
When executing `mvn clean install` both unit and integration tests will be executed after the packaging phase of the jar

Do not add these annotations to unit tests because the `@SpringBootTest` will add a context and deploy the whole springboot app to run the test which is slow (and also unit test do not need the whole springboot app)

I also refactored some test slightly to incorporate the changes.

Please do not merge this PR before #220 

# Reviewer(s)
 - @DPigeon 
 - @wltrf 
 - @abdulmansour 
 - @EspressoCode


razine-bensari:
The CI fails because in order to run the integration tests, I cannot to a DB at localhost, I assume there is no DB at local host at the moment the CI runs the integration tests. We need a definite endpoint for database. @DPigeon  @wltrf  based on previous conversation, if the VMs are full capacity, I will make another VM just for the DB, is that an acceptable solution?

DPigeon:
> The CI fails because in order to run the integration tests, I cannot to a DB at localhost, I assume there is no DB at local host at the moment the CI runs the integration tests. We need a definite endpoint for database. @DPigeon @wltrf based on previous conversation, if the VMs are full capacity, I will make another VM just for the DB, is that an acceptable solution?

I think we can use http://dev.money-tree.tech:7474/browser/ for that.

DPigeon:
Hey Razine I left a comment on the issue #221 . I want to make sure everything is fine between the two branches.

backend/src/main/resources/application.properties
```diff
@@ -1,11 +1,11 @@
-spring.profiles.active=dev
+spring.profiles.active=@spring.profiles.active@
 
 # REST related config
 server.servlet.context-path = /api/v1
 
 server.port=8080
 
-spring.data.neo4j.uri=bolt://localhost:7687
+spring.data.neo4j.uri=bolt://dev.money-tree.tech:7687
```
DPigeon, line 8: Here we could use the container`s name since we`re using docker compose. But the dev endpoinr will still work.

DPigeon (approved):

PR #236, opened 2021-01-12
Update pom.xml

razine-bensari:
# Related Issue
- upgrading library version that uses a string type for date instead of a lcoalDateTime type

# Reviewer(s)
 - @Alessandro100


Alessandro100 (approved):
nice

Issue #175, opened 2020-11-26
Add error handling object in the frontend


Alessandro100:
As a developer I would like a consistent error model to better display errors

**Acceptance Criteria**
- An error model is created in the frontend that mirrors the backend error model
- The model is implemented

razine-bensari:
This is what the error json looks like from backend api to front endL:

`{
  "status": "BAD_REQUEST",
  "timestamp": null,
  "message": "Illegal argument(s) in the URI",
  "debugMessage": "Invalid address format `bolt://IP_REMOTE_NEO4J:7687`"
}`

razine-bensari:
`{
  "status": "BAD_REQUEST",
  "timestamp": null,
  "message": "Illegal argument(s) in the URI",
  "debugMessage": "Invalid address format bolt://IP_REMOTE_NEO4J:7687"
}`

razine-bensari:
<img width="520" alt="Screen Shot 2021-01-09 at 12 37 02 PM" src="https://user-images.githubusercontent.com/34013446/104104636-60969800-5277-11eb-9120-f4d3918a85bd.png">

PR #237, opened 2021-01-13
Enhancement/#175 add error handling object in the frontend

EspressoCode:
# Related Issue
- #175 add error handling object in the frontend

# Proposed changes
- added a model for handling error messages in frontend which mirrors the error from backend

### DIFF
--- a/client/src/app/store/effects/app.effects.ts
+++ b/client/src/app/store/effects/app.effects.ts
@@ -7,6 +7,7 @@ import { TransactionService } from '../../services/transaction/transaction.servi
 import { UserService } from '../../services/user/user.service';
 import * as appActions from '../actions/app.actions';
 import { catchError, map, switchMap } from 'rxjs/operators';
+import { AppError } from 'src/app/interfaces/app-error';
 
 @Injectable()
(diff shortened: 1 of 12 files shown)


client/src/app/store/effects/app.effects.ts
```diff
@@ -80,10 +100,24 @@ export class Effects {
...
     )
   );
 }
+
+function mirrorError(backEndError): AppError {
+  const errorMessage: AppError = {
+    status: backEndError.error.status,
+    timestamp: backEndError.error.timestamp,
+    debugMessage: backEndError.error.debugMessage,
+    message: backEndError.error.message,
```
Alessandro100, line 120: As a defensive programming technique, we always need to assume that there is a possibility a chain of objects doesn't exist.

if (backendError && backendError.error) then ... else return null

Also, this is minor but can you change 'backEndError' to 'backendError'

Alessandro100 (changes requested):
I left a small comment, but otherwise it looks good! I think our errors will be much better handled this way

Alessandro100 (approved):
nice changes!

PR #239, opened 2021-01-14
Issue/#176 default user value

abdulmansour:
# Related Issue
- [link to the issue on zenhub](https://app.zenhub.com/workspaces/moneytree-board-5f5bcf1c7ea48f001c28c973/issues/dpigeon/money-tree/176)

# Proposed changes
- newly created user will have a predefined avatarURL pointing to a default image stored in s3
- when the user updates his default image, the default image is NOT deleted from s3
- endpoint exclusively used for updating image, change to => PUT /users/profile-picture/{id}

# Reviewer(s)
 - @razine-bensari @DPigeon @Alessandro100

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
+++ b/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
@@ -29,6 +29,7 @@ import com.capstone.moneytree.validator.ValidatorFactory;
 public class DefaultUserService implements UserService {
 
    private static final String USER_NOT_FOUND = "The requested user was not found";
+   private static final String DEFAULT_PROFILE_NAME = "DEFAULT-profile.jpg";
 
    private final UserDao userDao;
    private final ValidatorFactory validatorFactory;
@@ -37,9 +38,6 @@ public class DefaultUserService implements UserService {
    private final AmazonS3Service amazonS3Service;
 
    private final String bucketName;
-
-   @Value("${aws.default.profile.picture}")
-   private String defaultProfileURL;
 
    @Autowired
    public DefaultUserService(UserDao userDao, ValidatorFactory validatorFactory, AmazonS3Service amazonS3Service, @Value("${aws.profile.pictures.bucket}") String bucketName) {
@@ -81,7 +79,7 @@ public class DefaultUserService implements UserService {
       String password = user.getPassword();
       String encryptedPassword = encryptData(password);
       user.setPassword(encryptedPassword);
-      user.setAvatarURL(String.format("https://%s.s3.amazonaws.com/%s", bucketName, defaultProfileURL));
+      user.setAvatarURL(String.format("https://%s.s3.amazonaws.com/%s", bucketName, DEFAULT_PROFILE_NAME));
 
       userDao.save(user);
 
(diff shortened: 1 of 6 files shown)


Alessandro100 (changes requested):
This PR is going from your branch into master, please change it so it goes into the dev branch instead, after that this feature can be properly reviewed (right now there are 143 files changed)

backend/src/main/resources/application-local.properties
```diff
@@ -25,4 +25,5 @@ IEXCloud.secret.token=${IEXCLOUD_SECRET_TOKEN_SANDBOX}
 # Amazon S3 related properties
 aws.access.key.id=${AWS_ACCESS_KEY}
 aws.secret.access.key=${AWS_SECRET_ACCESS_KEY}
-aws.profile.pictures.bucket=${AWS_PROFILE_PICTURES_BUCKET}
\ No newline at end of file
+aws.profile.pictures.bucket=${AWS_PROFILE_PICTURES_BUCKET}
+aws.default.profile.picture=${DEFAULT_PROFILE_PICTURE}
```
razine-bensari, line 29: You do not need to store it as env variable, way too much work. The url does not need to be protected.
having `aws.default.profile.picture=https://my.default.url` is good enough

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -37,14 +37,16 @@
    private final AmazonS3Service amazonS3Service;
 
    private final String bucketName;
+   private final String defaultProfileURL;
 
    @Autowired
-   public DefaultUserService(UserDao userDao, ValidatorFactory validatorFactory, AmazonS3Service amazonS3Service, @Value("${aws.profile.pictures.bucket}") String bucketName) {
+   public DefaultUserService(UserDao userDao, ValidatorFactory validatorFactory, AmazonS3Service amazonS3Service, @Value("${aws.profile.pictures.bucket}") String bucketName, @Value("${aws.default.profile.picture}") String defaultProfileURL) {
       this.userDao = userDao;
```
razine-bensari, line 44: It does not make sense to pass a default url as constructor. You can inject it within the field attribute.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -37,14 +37,16 @@
    private final AmazonS3Service amazonS3Service;
 
    private final String bucketName;
+   private final String defaultProfileURL;
```
razine-bensari, line 40:
```
@Value("${aws.profile.pictures.bucket}")
private final String defaultProfileURL;
```

razine-bensari (changes requested):
Please update the default service constructor.

Also I think there will be some conflicts if David's brnach is merged first regarding generic user endpoint

backend/src/main/resources/application-local.properties
abdulmansour: sounds good

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
abdulmansour: right agreed, was coded at 2am 🤦‍♂️

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
abdulmansour: this relates to the previous comment, will make the adjustment

razine-bensari (approved):
Nice work!, Just make sure it builds green before merging

Alessandro100 (approved):
nice changes

PR #240, opened 2021-01-14
Issue/#174 generic update endpoint

abdulmansour:
# Related Issue
- [link to the issue on zenhub](https://app.zenhub.com/workspaces/moneytree-board-5f5bcf1c7ea48f001c28c973/issues/dpigeon/money-tree/174)

# Proposed changes
- updated endpoint to edit user profile properties PUT /users/{id} for generic update
- endpoint expects user id to be provided as path variable, and a User object in requestBody

# Reviewer(s)
 - @razine-bensari @DPigeon @Alessandro100


Alessandro100 (changes requested):
This PR is going from your branch into Mater, please change it to go to dev (I don't think you even need to make a new PR, just adjust stuff). As of right now I see 143 files changed (because it's comparing it to master) but there is also a frontend component to this task that I will have to refactor

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -81,11 +81,23 @@ public User createUser(User user) {
...
       userDao.save(user);
 
-      LOG.info("Created user: {}", user.getFirstName());
+      LOG.info("Created user: {}", user.getUsername());
 
       return user;
    }
 
+   @Override
+   public User updateUser(User userToUpdate, User user) {
```
razine-bensari, line 90: this endpoint shouold not have two user object. only one. we get the id of that user, call the userDao to fetch the user from the db, if the user exist then we validate the newly sent user and then update otherwise we return an error if the user is not valid. we also return an error there is no such user in the database

razine-bensari (changes requested):

abdulmansour:
@razine-bensari @Alessandro100 

checkout the branch, should now work to update the user by only providing the method that are of interest

Alessandro100 (approved):
I updated the frontend code to match the new backend and it works!

Issue #198, opened 2020-12-09
[BUG] Unique Email/Username Validation


wltrf:
**Describe the bug**
The /users/create-user endpoint doesn't check if an email is unique or if a username is unique

**To Reproduce**
PREREQ: A local backend/db combo running locally
1. Open Terminal and run the following command (or run in postman):
2. curl --location --request POST 'http://localhost:8080/api/v1/users/create-user' \
--header 'Content-Type: application/json' \
--data-raw '{
    "email": "hello.friend@gmail.com",
    "username": "hello",
    "password": "password123",
    "firstName": "hello",
    "lastName": "friend"
}'
3. curl --location --request POST 'http://localhost:8080/api/v1/users/create-user' \
--header 'Content-Type: application/json' \
--data-raw '{
    "email": "hello.friend@gmail.com",
    "username": "hello2",
    "password": "password123",
    "firstName": "hello",
    "lastName": "friend"
}'
4. curl --location --request POST 'http://localhost:8080/api/v1/users/create-user' \
--header 'Content-Type: application/json' \
--data-raw '{
    "email": "hello.friend2@gmail.com",
    "username": "hello",
    "password": "password123",
    "firstName": "hello",
    "lastName": "friend"
}'
5. See error (step 3 and 4 return ok). Alternatively, browse to http://localhost:7474/browser and see the three users created

**Expected behavior**
Step 3 and 4 should return errors since they use a non unique email and non unique username respectively.

**Desktop (please complete the following information):**
 - OS:
![Screenshot from 2020-12-09 10-39-08](https://user-images.githubusercontent.com/33789241/101653913-e40d6100-3a0d-11eb-88c3-0e52b344941c.png)
- $ curl --version
curl 7.68.0 (x86_64-pc-linux-gnu) libcurl/7.68.0 OpenSSL/1.1.1f zlib/1.2.11 brotli/1.0.7 libidn2/2.2.0 libpsl/0.21.0 (+libidn2/2.2.0) libssh/0.9.3/openssl/zlib nghttp2/1.40.0 librtmp/2.3
Release-Date: 2020-01-08
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp 
[...]

PR #241, opened 2021-01-14
#198 changed BE to check for username or email

EspressoCode:
# Related Issue
- This PR is a bug fix for #198 Unique-Email-Username-Validation 
# Proposed changes
- Changed the userAlreadyExists method to check for username or email separately.
- Now when the user tries to put a username or email that is already in database (for signUp), an error will be shown which specifies which one must be fixed.

# Reviewer(s)
@DPigeon  @razine-bensari  @Alessandro100


razine-bensari (approved):
Nice work!
Good refactoring on the test

Issue #179, opened 2020-11-27
Investigate / Optimize IEX Cloud API calls


Alessandro100:
As a developer I would like to optimize the IEX Cloud endpoint in the backend so I can make more calls

**Acceptance Criteria**
- Find a solution to reduce the amount of messages used in the IEX Cloud

<br>
<br>

Potential Solutions / Discussion Notes

1. ~~Take out a mortgage and pay for IEXCloud~~
2. Mock api/service as proposed by lindsay (wiremock)
3. Some sort of Proxy server
4. Store stock info in database as a mean of caching.
5. Not use /batch calls which are expensive

**Aditionnal Optimizations**
Use cache when making calls, from client to backend and/or from backend to iexcloud
https://web.dev/cache-api-quick-guide/
https://spring.io/guides/gs/caching/

The one below is paid, but we could implement something similar.
https://iexcloud.io/cloud-cache/

wltrf:
**Potential Solutions / Discussion Notes**

1. ~~Take out a mortgage and pay for IEXCloud~~
2. Mock api/service as proposed by lindsay
3. Some sort of Proxy server
4. Store stock info in database as a mean of caching.
5. Not use /batch calls which are expensive

**Aditionnal Optimizations**
Use cache when making calls, from client to backend and/or from backend to iexcloud
https://web.dev/cache-api-quick-guide/
https://spring.io/guides/gs/caching/

The one below is paid, but we could implement something similar.
https://iexcloud.io/cloud-cache/

wltrf:
https://query1.finance.yahoo.com/v8/finance/chart/AAPL

wltrf:
https://github.com/iexg/IEX-API/issues/718
^ issue from a guy discussing price issue for his app. The IEXCloud guys suggest to use some sort of caching (multiple solutions vaguely stated)

for insight only, not very useful

wltrf:
https://iexcloud.io/docs/api/#company
https://iexcloud.io/docs/api/#quote

Alessandro100:
Solution: Remove expensive calls and we'll deal with historical data later through Yahoo finance API

PR #245, opened 2021-01-17
Remove expensive calls from batch in facade

wltrf:
# Related Issue
- #179 

# Proposed changes
- Remove expensive calls for /batch in the facade (NEWS weight 10 and CHART/HISTORICAL PRICES weight 100)

# Additional Info
- This reduces the /batch/{symbol} call cost from 118 to 8, which is a good improvement.
- Further solution would be to have every user provide their own api key just like with alpaca (we could create it on their behalf)

# Reviewer(s)
 - @razine-bensari 
 - @abdulmansour 
 - @DPigeon

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/facade/StockMarketDataFacade.java
+++ b/backend/src/main/java/com/capstone/moneytree/facade/StockMarketDataFacade.java
@@ -60,10 +60,8 @@ public class StockMarketDataFacade {
                 .withSymbol(symbol)
                 .addType(BatchStocksType.BOOK)
                 .addType(BatchStocksType.COMPANY)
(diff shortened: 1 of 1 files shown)


razine-bensari (approved):
Lgtm

DPigeon (approved):
Looks good. We should have a meeting with what you mentioned in the pull request about having users generate their own IEXCloud key.

Alessandro100:
Nice changes! In terms of users bringing in their own IEX Cloud API key, I don't really see how we can manage that without ruining the user experience. Alpaca uses OAuth to authenticate functionality on their site, IEX Cloud just provides data. In a future where if this application is used by many people, we would just increase our IEX Cloud plan.

Issue #226, opened 2021-01-11
Buy/sell UI elements


Alessandro100:
As a user I would like to see the buy sell elements on the stock page so I know what actions I can do

Note:
- this story is only to implement the elements in the pictures without functionality. The popups will be another story

**Acceptance Criteria**
- The designs are identical to the Figma designs
- The objects are added in the stock detail page
- Both states are taken care of

This is if the user owns stock
![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/608d6d84-e099-430d-943b-41417dfa8464)

This is if they don't
![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/3e61ef57-8a9c-4409-93a3-8d5bbf3dab6f)

PR #247, opened 2021-01-19
Issue/#226 buy sell UI

marwan-ayadi:
# Related Issue
- #226 

# Proposed changes
- add buy/sell ui elements in stock-detail page 

# Reviewer(s)
 - @Alessandro100


client/src/index.html
```diff
@@ -1,15 +1,21 @@
...
+    <link
+      href="https://fonts.googleapis.com/css?family=Roboto:300,400,500&display=swap"
+      rel="stylesheet"
+    />
+    <link
+      href="https://fonts.googleapis.com/icon?family=Material+Icons"
+      rel="stylesheet"
+    />
+  </head>
+  <body class="mat-typography" style="background-color: #f8f8fb">
```
Alessandro100, line 18: this is bad css practice, we should put this style property in styles.scss file

Alessandro100 (changes requested):
It looks good, but there are a couple of things missing! 
image
Here if the total value is positive, the text should be green (red if its negative).

Secondly, in the story it also asks to make the UI if the user doesn't own any stocks. Check the issue and you'll see what I mean.

I also left a small css change

client/src/index.html
marwan-ayadi: resolved

Alessandro100 (approved):
Nice changes

PR #249, opened 2021-01-21
First release for production (Post release 1)

Alessandro100:
This is the first release to production to get user feedback

# Proposed changes
User is able to login, search for a stock and view stock information

# Checklist (if applicable)
- [X] Tests
- [X] Documentation

# Reviewer(s)
 - @DPigeon @razine-bensari


DPigeon:
So we have about 22 features and 10 fixes (counted from issues and merges in dev). To have a great version, we should have 22 commits with title 'feat(something)' and 10 commits with title 'fix(something)'. If we do that, our app version will be bumped from v0.0.3 to v0.22.10. There must be a faster way to commit in total 32 messages like that for the version to show up in our semantic versioning system. We'll have to look into that or forget about it.

DPigeon:
The 502 bad gateway has been fixed. I think now we're ready to deploy to prod.
@Alessandro100 I'll approve the PR but make sure you do the versioning part. I'll let you handle that. If you need help ping me.

DPigeon (approved):

Issue #253, opened 2021-01-22
Add Certificates in Secrets


DPigeon:
As a developer, I want to add certificates into the secrets in the repo and use them in the CICD pipeline so that nobody can have access to them.

To achieve this, it is important to add the 4 files into the secrets of the repo. Then in the CICD pipeline, we need a way to inject these 4 secrets and create a file for each of them, and append those into the uat/prod build when deployed.

**Acceptance of criteria**:
- [x] Generate new private keys
- [x] Add the 4 secrets into the repo
- [x] Remove the certs directory in client
- [x] Write a script that creates the appropriate cert files depending on which branch CICD is
- [x] Add the files into the build of the deployment with a docker volume for example

wltrf:
i think this could be simplified using pkcs12 (password protected cert and key in one file). Not perfect in terms of security (prone to brute force attacks), but with a strong enough password it should be good enough for our proof of concept. So youd have the .p12 file residing on github and a password as a secret/env var

DPigeon:
> i think this could be simplified using pkcs12 (password protected cert and key in one file). Not perfect in terms of security (prone to brute force attacks), but with a strong enough password it should be good enough for our proof of concept. So youd have the .p12 file residing on github and a password as a secret/env var

Cool idea! Is this the same thing as having both dev and prod keys in secrets and then having both certificates on GitHub?

DPigeon:
I'll begin this story today.

PR #256, opened 2021-01-23
#253: Add Certificates in Secrets

DPigeon:
# Related Issue
- #253 

# Proposed changes
- Added certificates and keys into secrets
- CD uses the secrets to pass them into the angular dockerfile
- A script is adding them to the proper environment depending on the deployment

# Additional Info
- Since certificates were not hidden, a flow has been created for them to be hidden using secrets in our repo.

# Reviewer(s)
 - @wltrf 
 - @razine-bensari 
 - @abdulmansour


client/Dockerfile
```diff
@@ -1,7 +1,9 @@
 # Client Build
 FROM trion/ng-cli as ng
-# Define the env config arg
+# Define args
 ARG ENVCONFIG
+ARG CERT
+ARG CERT_KEY
```
wltrf, line 6: Should we really package the certificate and key inside the docker image? This exposes the private key and requires us to rebuild the image everytime we want to change the cert/key. I think it should simply be an env variable on the vm.

client/certs/money-tree.tech/cert.pem
```diff
@@ -1,56 +0,0 @@
...
-Y3J5cHQub3JnMDwGA1UdHwQ1MDMwMaAvoC2GK2h0dHA6Ly9jcmwuaWRlbnRydXN0
-LmNvbS9EU1RST09UQ0FYM0NSTC5jcmwwHQYDVR0OBBYEFBQusxe3WFbLrlAJQOYf
-r52LFMLGMB0GA1UdJQQWMBQGCCsGAQUFBwMBBggrBgEFBQcDAjANBgkqhkiG9w0B
-AQsFAAOCAQEA2UzgyfWEiDcx27sT4rP8i2tiEmxYt0l+PAK3qB8oYevO4C5z70kH
-ejWEHx2taPDY/laBL21/WKZuNTYQHHPD5b1tXgHXbnL7KqC401dk5VvCadTQsvd8
-S8MXjohyc9z9/G2948kLjmE6Flh9dDYrVYA9x2O+hEPGOaEOa1eePynBgPayvUfL
-qjBstzLhWVQLGAkXXmNs+5ZnPBxzDJOLxhF2JIbeQAcH5H0tZrUlo5ZYyOqA7s9p
-O5b85o3AM/OJ+CktFBQtfvBhcJVd9wvlwPsk+uyOy2HI7mNxKKgsBTt375teA2Tw
-UdHkhVNcsAKX1H7GNNLOEADksd86wuoXvg==
------END CERTIFICATE-----
```
wltrf, line 56: you could have kept the certificate here, this will be public anyway.

wltrf (changes requested):

client/Dockerfile
DPigeon: Yeah you're right. What I am thinking right now is having the keys and certificates in their respective VM and then mounting a volume to the container when we start docker compose. I think this is a valid solution and all the scripting/secrets can be removed with that. As soon as we start docker compose, the client service will mount the certificates to the ./ssl directory on the container and nginx will be able to pick them up easily.

client/certs/money-tree.tech/cert.pem
DPigeon: I think I'll remove them anyway just for security reasons. Let's keep everything in the VM.

razine-bensari (approved):
Good work

wltrf (approved):

Issue #228, opened 2021-01-11
Sell Stock popup UI element


Alessandro100:
As a user I would like to see all the information of a buy transaction before I sell

**Acceptance Criteria**
- Design matches the figma design
- Functionality matches the steps under

Step 1: choose selling type (market or limit)
![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/02bacba9-acc0-493e-a851-ceee732143bb)

Step 2a: if market display
![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/de0e60b4-0d3a-49f3-809c-c466ae4bf187)

Step 2b: if limit display
![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/c447b90a-6cc8-4fb8-9e59-10ef8abe4cce)

PR #260, opened 2021-01-25
Feature/#228 sell stock UI

marwan-ayadi:
# Related Issue
#228 

# Proposed changes
- Add Sell Actions popup ui 

# Reviewer(s)
 - @Alessandro100

### DIFF
--- a/client/src/theme.scss
+++ b/client/src/theme.scss
@@ -122,12 +122,12 @@ $mt-theme: mat-light-theme($mt-palette-primary, $mt-palette-accent);
   border-radius: 5px;
 }
 
-@media only screen and (max-width: 600px) {
-  ::ng-deep .cdk-overlay-pane {
-    max-width: none !important;
-    height: 100%;
-  }
-  ::ng-deep .mat-dialog-container {
-    padding: 0px;
-  }
-}
+// @media only screen and (max-width: 600px) {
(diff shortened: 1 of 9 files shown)


client/src/theme.scss
```diff
@@ -122,12 +122,12 @@ $mt-theme: mat-light-theme($mt-palette-primary, $mt-palette-accent);
...
-}
+// @media only screen and (max-width: 600px) {
+//   ::ng-deep .cdk-overlay-pane {
+//     max-width: none !important;
+//     height: 100%;
+//   }
+//   ::ng-deep .mat-dialog-container {
+//     padding: 0px;
+//   }
+// }
```
Alessandro100, line 133: We need this for other popups that appear in the app such as edit user. Note, this only affects the mobile version which should have the same behaviour.

Alessandro100 (commented):
The component looks good, I left one comment about the code you removed. Once that is addressed I can approve it

client/src/theme.scss
marwan-ayadi: It actually not good for mobile and certainly not good to use ng-deep. It would be better to change the dialog in the edit-user component and not for all components.

Alessandro100 (approved):
Nice changes, it looks really good

Issue #182, opened 2020-11-27
Create edit profile page form logic


Alessandro100:
As a developer I would like to edit my profile to make sure my profile is up to date

**Acceptance Criteria**
- Ability to edit Name
- Ability to edit picture
- Ability to edit background image
- Ability to edit password
- Name cannot be empty (validation)
- Must set the new user data in the NgRx

![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/14d9404f-8589-4a53-ae92-f857b63e48fe)

PR #281, opened 2021-01-27
Feature/#182 create edit profile page form logic

EspressoCode:
# Related Issue
- #182 create edit profile page form logic

# Proposed changes
- Added logic to edit profile component and the dialog in the home page, using the two backend endpoints for user photo change and the general update user, a user can edit profile info as well as the photo.

# Additional Info
- I made this pr for a high-level review of how the logic was implemented, so there will be a couple of changes and unit/e2e tests for the components which will be added in a future commit.

# Reviewer(s)
@Alessandro100

### DIFF
--- a/client/src/app/components/edit-profile/edit-profile.component.ts
+++ b/client/src/app/components/edit-profile/edit-profile.component.ts
@@ -1,5 +1,4 @@
-import { Component, OnInit, EventEmitter, Input, Output } from '@angular/core';
-
+import { Component, Inject } from '@angular/core';
 import {
   FormBuilder,
   FormGroup,
@@ -7,14 +6,8 @@ import {
   AbstractControl,
 } from '@angular/forms';
 import { User } from 'src/app/interfaces/user';
-
-// export interface ProfileData {
-//   firstName: string;
-//   lastName: string;
-//   password: string;
-//   biography: string;
-// }
-
+import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { ProfileData } from '../../interfaces/profileData';
 @Component({
   selector: 'app-edit-profile',
   templateUrl: './edit-profile.component.html',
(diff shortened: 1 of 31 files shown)


client/src/app/components/edit-profile/edit-profile.component.ts
```diff
@@ -1,22 +1,234 @@
...
+    return false;
+  }
+  sanitizeImageUrl(imageUrl: string): SafeUrl {
+    // otherwise the browser will complaint about unsafe url
+    return this.sanitizer.bypassSecurityTrustUrl(imageUrl);
+  }
+  cancelPhoto(): void {
+    this.temporaryPhotoFile = null;
+    this.userPhotoURL = this.userInState.avatarURL;
+    console.log('remove photo clicked');
```
Alessandro100, line 226: remove this console log

Alessandro100 (changes requested):
Very nicely implemented with NgRx, the code looks good. There are a couple of issues. For the picture, the user can choose any file type, we can force them to choose only image tags. Second thing is that the cover image functionality doesn't work

EspressoCode:
> Very nicely implemented with NgRx, the code looks good. There are a couple of issues. For the picture, the user can choose any file type, we can force them to choose only image tags. Second thing is that the cover image functionality doesn't work
I think I implemented that, so if you try it with a pdf file for example, you should see the error.
in the ts file:
`if (type !== 'image/jpeg' && type !== 'image/png') {
      this.pictureErrMessage =
        'The photo must be a file of type: jpeg, png, jpg!';
      this.userPhotoURL = this.userInState.avatarURL;
      this.temporaryPhotoFile = null;
      return;}`
in template file:
`<div class="profile-picture">
      profile picture
      <button mat-icon-button class="edit-profile-pic" *ngIf="userPhotoURL==userInState.avatarURL">
        <input single type="file" id="profile-picture" style="display: none;" **accept="image/x-png,image/jpeg"**
          (change)="onFileSelected($event)" />
        <label for="profile-picture">
          <mat-icon>add_a_photo</mat-icon>
        </label>
      </button>`

EspressoCode:
In the commit before the last commit (3f61fde), I added cover photo to both the frontend and backend. This needs to be reviewed by the backend to make sure we can safely merge. Tests for backends also fixed based on new changes, so please review this pr and especially the last commit, thanks!
@razine-bensari @DPigeon @wltrf @abdulmansour

Alessandro100 (approved):
Nice job on the story! Good initiative on the backend section, and the tests look really good.

Issue #287, opened 2021-02-04
Backend Properties Not Set Properly


DPigeon:
**Describe the bug**
When running the backend locally, the dev profile is loaded.

**To Reproduce**
1. Run the backend
2. It loads dev profile
3. An error is obtained saying it is trying to fetch the dev database

**Expected behavior**
It should load the local properties when running locally (in application.properties) and load other profiles depending on the deployment type.

Bug reported by @DPigeon

PR #288, opened 2021-02-04
#287: Fix Backend Local & Dev Properties

DPigeon:
# Related Issue
- #287 

# Proposed changes
- Local properties are not loaded by default
- Other properties such as dev and prod can still be loaded when deploying with profile to profile=dev or profile=prod

# Reviewer(s)
 - @razine-bensari 
 - @Alessandro100


razine-bensari (approved):
lgtm

Issue #276, opened 2021-01-26
Availability of the stock market


Alessandro100:
As a user I would like to know if the stock market is open in a very easy to see manner

Discussion:
- Since we only support NYSE, NASDAQ, and AMEX which all open and close 9:30am-4pm (eastern) do you think it's worth not even having an API call for this and that we can do a simple date / time check in the frontend?

Note:
- We do not have a figma design for when the market is closed. We can either remove the "open market" token or we can replace it with a token that says "closed market" in a gray /red token box

**Acceptance Criteria**
- The UI is implemented for when the market is opened / when the market is closed
- The functionality of the time is implemented
- The calendar endpoint is implemented 

![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/3de8173e-fd95-41df-9070-d9e967c5d191)

Alessandro100:
https://alpaca.markets/docs/api-documentation/api-v2/calendar/

PR #291, opened 2021-02-09
Feature/#276 availability of the stock market

EspressoCode:
# Related Issue
- #276 Availability of the stock market

# Proposed changes
Backend:
Added a new endpoint to get the Alpaca market clock.
Added new Exception for that specific endpoint.
Fixed a bug where the parameters for the request header were in the wrong order and causing the request to respond with 401.

Frontend: 
added MarketClock interface
market clock added to State
some styling improvements for stock header 
added a line where the user can see the next opening time of the market when the market is closed.

### DIFF
--- a/client/src/app/components/stock-detail-header/stock-detail-header.component.ts
+++ b/client/src/app/components/stock-detail-header/stock-detail-header.component.ts
@@ -1,6 +1,6 @@
+import { MarketClock } from './../../interfaces/market-clock';
 import { Component, Input } from '@angular/core';
 import { Stock } from './../../interfaces/stock';
-import { StockService } from '../../services/stock/stock.service';
 
 @Component({
   selector: 'app-stock-detail-header',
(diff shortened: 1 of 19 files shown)


client/src/app/components/stock-detail-header/stock-detail-header.component.ts
```diff
@@ -51,4 +54,10 @@ export class StockDetailHeaderComponent {
     }
     return '';
   }
+  nextOpen(): string {
+    if (this.marketClock) {
+      return this.marketClock.nextOpen.replace('T', ' at ').split(':00-')[0];
+    }
+    return '';
```
Alessandro100, line 61: this could be written as `return this.marketClock ? this.marketClock.... : ''`

client/src/app/store/effects/app.effects.ts
```diff
@@ -35,6 +35,32 @@ export class Effects {
...
+      switchMap((action) => {
+        return this.stockService.loadMarketClock().pipe(
+          map((data: any) =>
+            appActions.loadMarketClockSuccess({
+              marketClock: {
+                isOpen: data.body.isOpen,
+                nextClose: data.body.nextClose,
+                nextOpen: data.body.nextOpen,
+                timestamp: data.body.timestamp,
+              },
```
Alessandro100, line 50: we can have a separate function similar to 'formatUser' so the formatting logic is separate

Alessandro100 (changes requested):
Really nice job! Your code looks pretty good, I made a few refactoring comments. Another thing is that since this is an acceptance test story, you will need to write an system test to prove that the feature works. Otherwise, great use of interfaces, NgRx, and backend code

Alessandro100 (approved):
Nice changes

Issue #271, opened 2021-01-26
Create UI elements for Following (Home Page)


Alessandro100:
As a user I would like to see who I am following and their scores

**Acceptance Criteria**
- The UI elements are placed exactly like in the picture
- The UI is placed in the home page

![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/fbc87c82-5d0e-4b79-9fe5-02d512ca8db0)

PR #293, opened 2021-02-10
Feature/#271 followings UI

marwan-ayadi:
# Related Issue
#271 

# Proposed changes
- Reorganize home page layout
- Add Mat-card with the following list (name, profile pic, score)
- Add search input to filter following list
- following name allow a user to navigate to the following profile

# Reviewer(s)
 - @Alessandro100

### DIFF
--- a/client/src/app/components/followings-search/followings-search.component.html
+++ b/client/src/app/components/followings-search/followings-search.component.html
@@ -5,7 +5,6 @@
       <mat-icon>search</mat-icon>
     </button>
     <input
-      class="search-following"
       [(ngModel)]="followingSearch"
       matInput
       placeholder="Search for followings"
--- a/client/src/app/components/followings-search/followings-search.component.html
+++ b/client/src/app/components/followings-search/followings-search.component.html
@@ -1,10 +1,26 @@
 <mat-card class="followings-container">
-  <h3>Followings ({{ followings.length }})</h3>
-  <input
-    class="followings-search"
-    [(ngModel)]="followingSearch"
-    matInput
-    placeholder="Search for followings" />
+  <h3 class="title">Followings ({{ followings.length }})</h3>
+  <mat-form-field appearance="outline">
+    <button class="search-icon" mat-icon-button matPrefix disabled>
+      <mat-icon>search</mat-icon>
+    </button>
+    <input
+      class="search-following"
+      [(ngModel)]="followingSearch"
+      matInput
+      placeholder="Search for followings"
+    />
+    <button
+      class="clear-icon"
+      mat-icon-button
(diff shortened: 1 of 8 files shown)


client/src/app/components/followings-search/followings-search.component.html
```diff
@@ -0,0 +1,37 @@
...
+          width="40"
+          height="40"
+          [src]="
+            'https://static.wikia.nocookie.net/bleachuntoldstories/images/2/24/Hitsugaya_toshiro_profile.png'
+          "
+        />
+        <div class="following-info">
+          <span
+            class="following-name"
+            (click)="navigateToFollowingProfile(following.firstName)"
```
EspressoCode, line 23: I know that this will be modified in future because we might have 2 followings with the same first name therefore we cannot use only the first name to define the route/ticker. I suggest adding a TODO comment here.

client/src/app/components/followings-search/followings-search.component.ts
```diff
@@ -0,0 +1,58 @@
...
+  ]; // to be changed
+  followingSearch = '';
+
+  constructor(private router: Router) {}
+  ngOnInit(): void {}
+
+  navigateToFollowingProfile(following: string): void {
+    this.router.navigate(['/user-profile/' + following]);
+  }
+  isSearchedFollowings(following): boolean {
```
EspressoCode, line 47: I tried the search bar and realized that if you put first name + space + a part of the last name it will not show any results (for instance "hossein n" will return no results), if you could add logic for that case it will be more convenient for users of our application.

EspressoCode (changes requested):

EspressoCode:
Very good job on the styling, it is pretty much the same as our Figma design. These are some minor changes that can be done to still improve it:
Following title should be bold, with more padding from left.
Padding for the search bar also should be increased from left and right.
Fonts color and size for the name of the following and the score should be changed.
Following

The result card will look better if the round corners become squre.
round borders

And the lines between followings should be removed:
lines

EspressoCode:
Very good job on this pr, I left you a couple of comments.

client/src/app/components/followings-search/followings-search.component.html
marwan-ayadi: yeah this will be modified when we will do the logic

client/src/app/components/followings-search/followings-search.component.ts
marwan-ayadi: for now the search is only a filter. When I will do the logic I may use the same search library that we use in stock search

EspressoCode (approved):

Issue #266, opened 2021-01-26
Add historical data to the stock endpoint


Alessandro100:
As a developer I would like to receive the historical stock data to see how the stock did in the past

Notes:
- When the get stock endpoint is called, it should return the historical data
- Ideally we wouldn't use IEXCloud for this as it's expensive
- @wltrf mentioned something about scraping it from yahoo, could be very interesting solution
- frontend: #267 

**Acceptance Criteria**
- When the frontend calls the GET stock endpoint, it should return all the stock information + the historical data ready to plot

PR #295, opened 2021-02-14
Implement Yahoo finance proxy for hist data #266

wltrf:
Implement Yahoo finance proxy for grap historical data

~~**READ BELOW**~~
~~I implemented 2 options to do this.~~
~~/chart1~~ /yahooChart uses a service/facade flow which looks clean and may be reused if we decide to get more data from yahoo finance.
~~/chart2 mirrors the request using a proxy. it is very short and simple.
Comments/Preferences?~~

# Related Issue
#266

# Proposed changes
- Add /yahooChart endpoint to the stockmarket controller
- Add yahoo service
- add yahoo facade

# Additional Info
- the endpoint esentially relays/proxy the response from https://query1.finance.yahoo.com/v8/finance/chart/{ticker}?range={range}&interval={interval}

# Checklist (if applicable)
- [x] Tests
- [x] Documentation

# Reviewer(s)
 - backend team

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/controller/StockController.java
+++ b/backend/src/main/java/com/capstone/moneytree/controller/StockController.java
@@ -102,17 +102,8 @@ public class StockController {
    Valid ranges: [1d, 5d, 1mo, 3mo, 6mo, 1y, 2y, 5y, 10y, ytd, max]
    Valid intervals: [1m, 2m, 5m, 15m, 30m, 60m, 90m, 1h, 1d, 5d, 1wk, 1mo, 3mo]
    */
-   @GetMapping(value = "/chart1/{symbol}", produces = {"application/json"})
-   public Mono<String> getChart1(@PathVariable String symbol, @RequestParam String range, @RequestParam String interval){
+   @GetMapping(value = "/yahooChart/{symbol}", produces = {"application/json"})
+   public Mono<String> getYahooChart(@PathVariable String symbol, @RequestParam String range, @RequestParam String interval){
       return yahooFinanceService.getHistoricalGraphData(symbol, range, interval);
-   }
-   
-   /*
-   Valid ranges: [1d, 5d, 1mo, 3mo, 6mo, 1y, 2y, 5y, 10y, ytd, max]
-   Valid intervals: [1m, 2m, 5m, 15m, 30m, 60m, 90m, 1h, 1d, 5d, 1wk, 1mo, 3mo]
-   */
-   @GetMapping(value = "/chart2/{symbol}", produces = {"application/json"})
-   public ResponseEntity<String> getChart2(ProxyExchange<String> proxy, @PathVariable String symbol, @RequestParam String range, @RequestParam String interval) {
-      return proxy.uri("https://query1.finance.yahoo.com/v8/finance/chart/" + symbol + "?range=" + range + "&interval=" + interval).get();
    }
 }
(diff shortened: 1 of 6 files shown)


DPigeon:
Option 1 looks very clean and better in my opinion. It looks like we'll have to use more data from Yahoo to overcome our IEXCloud API restriction problem. I'm on board with option 1.

backend/src/main/java/com/capstone/moneytree/controller/StockController.java
```diff
@@ -94,4 +92,13 @@ public StockController(StockMarketDataService stockMarketDataService) {
       Logo logo = stockMarketDataService.getLogo(symbol);
       return ResponseEntity.ok(logo);
    }
+
+   /*
+   Valid ranges: [1d, 5d, 1mo, 3mo, 6mo, 1y, 2y, 5y, 10y, ytd, max]
+   Valid intervals: [1m, 2m, 5m, 15m, 30m, 60m, 90m, 1h, 1d, 5d, 1wk, 1mo, 3mo]
+   */
+   @GetMapping(value = "/yahoochart/{symbol}", produces = {"application/json"})
```
DPigeon, line 100: Do we absolutely need ```produces = {"application/json"}``` here? I thought the output was JSON by default.

backend/src/main/java/com/capstone/moneytree/facade/YahooFinanceFacade.java
```diff
@@ -0,0 +1,22 @@
...
+import reactor.core.publisher.Mono;
+
+@Component
+public class YahooFinanceFacade {
+
+    private final WebClient webClient;
+
+    @Autowired
+    public YahooFinanceFacade(WebClient.Builder webClientBuilder) {
+        this.webClient = webClientBuilder.baseUrl("https://query1.finance.yahoo.com").build();
```
DPigeon, line 15: Refactor the URL string maybe in the properties or on top as a private final variable.

DPigeon (changes requested):
Nice work! Left some comments.

I think we should test 1-2 more paths in the IT test. Maybe an unsuccesful path test?
Also, like we talked in the meeting yesterday, we may need to add unit tests to this to test individual components. I know its a bit hard since we're using facades but we can always make basic unit tests on instance initialization and looking if the instance has been initialized properly (not null for example).

backend/src/main/java/com/capstone/moneytree/controller/StockController.java
wltrf: In this case yes, since it's returning a String (formatted as json), the content-type of the header is text/html so the explicit json header is required. 
Screenshot from 2021-02-19 19-18-12

backend/src/main/java/com/capstone/moneytree/facade/YahooFinanceFacade.java
wltrf: good idea

backend/src/main/java/com/capstone/moneytree/controller/StockController.java
DPigeon: Ahh I see. Thanks!

backend/src/main/java/com/capstone/moneytree/facade/YahooFinanceFacade.java
DPigeon: Fair enough.

DPigeon (approved):
Approved. Good work!

Issue #278, opened 2021-01-26
When an order is completed / fulfilled send an email to the user


Alessandro100:
As a user I would like to be notified if one of my orders is completed

**Acceptance Criteria**
- Send an email to the user when one of their orders is completed

Alessandro100:
Attributes in the email:
- stock name + symbol
- quantity purchased
- buy/sell action (market / limit)
- average share price
- timestamp
- total amount

Issue #277, opened 2021-01-26
Update user transactions and score based off webhooks trade updates


Alessandro100:
As a user I would like to be notified immediately about new trade updates

Question: Is the backend logic implemented in what the backend will return to the frontend?

**Acceptance Criteria**
- Once the user receives a trade update from the webhook it will update their: transactions, and score

PR #299, opened 2021-02-17
#278: Send Email When Is Order Fulfilled 

DPigeon:
# Related Issue
- #278 

# Proposed changes
- Send a simple email to the user receiving streams of filled order
- Unit Tests for fake SMTP server
- Integration Tests for email sending using a fake SMTP server
- Added new test properties used for special tests such as email ITs

# Additional Info
- Feel free to comment on the email format. I don't think it should be complex and it is not really a priority to create some HTML elements onto it.

# Checklist (if applicable)
- [x] Tests

# Screenshots (if applicable)
Title of the email seen for the user:
title

User receives an email like this one:
email3

# Reviewer(s)
 - @razine-bensari 
 - @abdulmansour 
 - @wltrf 
 - @Alessandro100


EspressoCode:
It looks very good, as a suggestion, could you please add a dollar sign ($) for the total amount and the average price?

backend/src/main/java/com/capstone/moneytree/facade/MarketInteractionsFacade.java
```diff
@@ -183,10 +188,30 @@ public void onStreamUpdate(AlpacaStreamMessageType streamMessageType, AlpacaStre
                TradeUpdateMessage tradeMessage = (TradeUpdateMessage) streamMessage;
                TradeUpdate tradeUpdate = tradeMessage.getData();
                if (tradeUpdate.getEvent().equals("fill")) {
```
EspressoCode, line 190: should this be "fulfilled" instead of "fill"?

backend/src/main/java/com/capstone/moneytree/facade/MarketInteractionsFacade.java
```diff
@@ -183,10 +188,30 @@ public void onStreamUpdate(AlpacaStreamMessageType streamMessageType, AlpacaStre
...
-                  messageSender.convertAndSend("/queue/user-" + userId, tradeUpdate.getOrder().getClientOrderId());
+                  messageSender.convertAndSend(
+                          "/queue/user-" + userId,
+                          tradeUpdate.getOrder().getClientOrderId());
+                  sendOrderCompletedEmail(userId, tradeUpdate);
```
EspressoCode, line 194: should we change the transaction status here? (from PENDING to FULFILLED/FILL/COMPLETE)?

DPigeon:
> It looks very good, as a suggestion, could you please add a dollar sign ($) for the total amount and the average price?
> 

Yes that is a good suggestion. I'll do it!

backend/src/main/java/com/capstone/moneytree/facade/EmailSender.java
```diff
@@ -0,0 +1,71 @@
...
+        return "Hello " + username + ",\n" +
+                "Thank you for trading with us. Your order details are indicated below.\n" +
+                "If you would like to view the status of your order, please visit http://money-tree.tech.\n\n" +
+                "Stock Purchased: " + stockName + "\n" +
+                "Quantity: " + quantity + "\n" +
+                "Average Share Price: " + avgPrice + "\n" +
+                "Time of Purchase: " + timestamp.getMonth().getValue() + "/" + timestamp.getDayOfMonth() + "/" + timestamp.getYear() +
+                " at " + timestamp.getHour() + ":" + timestamp.getMinute() + ":" + timestamp.getSecond() + "\n" +
+                "Total Amount: " + totalAmount + "\n" +
+                "Action: " + action + "\n\n" +
```
wltrf, line 67: This is just a suggestion: this could be in a class method (like toString()), but since we didn't write the Order class, I'm thinking maybe a modified version of the visitor design pattern (something like a helper class if you like). It would allow you to write orderVisitor.visit(order). In that visitor class, there could also be other methods for different printing structure like logVisit for logging or debugVisit for debugging.

wltrf (approved):
Good job.

backend/src/main/java/com/capstone/moneytree/facade/MarketInteractionsFacade.java
DPigeon: Hmm, I think so. Not sure though since it might have been done already by @razine-bensari and @abdulmansour in their story. 

@Alessandro100 what do you think? We can always do what @EspressoCode asked for in the current webhook build #277.

backend/src/main/java/com/capstone/moneytree/facade/EmailSender.java
DPigeon: Yeah I wanted to have a separate Mail class that contains all the necessary information for it. I'll see what I can do.

backend/src/main/java/com/capstone/moneytree/facade/MarketInteractionsFacade.java
DPigeon: Update: Let's do this in another story. This story is used to send an emaila and not update the transaction. I believe this will be done in the refactored story.

backend/src/main/java/com/capstone/moneytree/facade/EmailSender.java
DPigeon: Thanks for the comment Walter! I have created a new Mail class object and refactored everything inside it. Also created a toString() method that is used for the construction of the email.

razine-bensari (approved):
good work

backend/src/main/java/com/capstone/moneytree/facade/EmailSender.java
razine-bensari: That would be ideal for a refactoring activity!

wltrf: that's exactly what I thought when commenting this. Who would've thought refactoring would make us all excited :P.

Issue #273, opened 2021-01-26
Implement real data into followers


Alessandro100:
Note:
- [Discussion]: Will the "followers" data be attached to the user profile?

**Acceptance Criteria**
- Display correct follower information on the home page

![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/bfca8808-b6d6-4109-9c96-06f4613046da)

PR #306, opened 2021-02-22
Feature/#273 implement real data into followers

EspressoCode:
# Related Issue
#273 implement real data into followers

# Proposed changes
- Refactored backend endpoint implementation for followers
- Added followings (the list of users that are followed by the current user) endpoint in the backend
- Changed the relationship logic by adding followsDao and its functions to easily query on Follows relationships
- Added SanitizedUser class to return only necessary information about followers/followings users which is good for both security and performance
- Removed followers/followings from user fields and added separate objects to the state for followers/followings

# Reviewer(s)
 @Alessandro100 @DPigeon @razine-bensari @wltrf


backend/src/main/java/com/capstone/moneytree/model/SanitizedUser.java
```diff
@@ -0,0 +1,86 @@
+package com.capstone.moneytree.model;
+
+import com.capstone.moneytree.model.node.User;
+
+public class SanitizedUser { // this class will be used to return only the necessary info about a user for follower/followings api and other apis
```
DPigeon, line 5: To avoid duplicates, would this be better if we created a new constructor in the User model class with all the necessary parameters instead of creating a new class? And then you would get a SanitizedUser with a new getter as well only getting the needed info for the user.

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
```diff
@@ -123,10 +124,10 @@ public User login(@RequestBody User credentials) throws CredentialNotFoundExcept
     */
 
    @PostMapping("/{userId}/follow/{userToFollowId}")
-   ResponseEntity<Long> followUser(@PathVariable Long userId, @PathVariable Long userToFollowId) {
-      Long id = this.userService.followUser(userId, userToFollowId);
+   ResponseEntity<String> followUser(@PathVariable Long userId, @PathVariable Long userToFollowId) {
+      String res = this.userService.followUser(userId, userToFollowId);
```
DPigeon, line 128: What's the motivation with returning a String as a response with "userX is following userY"? Is it because the frontend needs to know who the current user is following explicitely? Normally, we only return relevant information in an object if needed in a response and not a string. Just curious.

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
```diff
@@ -137,10 +138,10 @@ public User login(@RequestBody User credentials) throws CredentialNotFoundExcept
     */
 
    @DeleteMapping("/{userId}/unfollow/{userToUnfollowId}")
-   ResponseEntity<Long> unfollowUser(@PathVariable Long userId, @PathVariable Long userToUnfollowId) {
-      Long id = this.userService.unfollowUser(userId, userToUnfollowId);
+   ResponseEntity<String> unfollowUser(@PathVariable Long userId, @PathVariable Long userToUnfollowId) {
+      String res = this.userService.unfollowUser(userId, userToUnfollowId);
```
DPigeon, line 142: Same as above here.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -258,31 +264,75 @@ public User login(User credentials)
...
+      }
+      Date currentDate = new Date();
+      Follows newFollowRel = new Follows(user, userToFollow, currentDate);
+      followsDao.save(newFollowRel);
 
-      user.follow(userToFollow);
-      userDao.save(user);
-
-      return userToFollowId;
+      return user.getUsername() + " followed " + userToFollow.getUsername();
```
DPigeon, line 287: Ahh I see now what we're giving to the frontend as a response. Maybe we could create a FollowResponse object containing the `user.getUsername()` and `userToFollow.getUsername()` inside? An object as such:
```
{
username: 'string'
follows: 'string'
}

That way, if we ever change anything on the frontend, then we don't have to change the backend as well. I feel like returning a string is prone to changes since we may have to edit the string format. Having an object gives the frontend a way to do whatever they want with the data. What do you think?
```

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -258,31 +264,75 @@ public User login(User credentials)
...
 
-      user.unfollow(userToUnfollow);
-      userDao.save(user);
+      List<Follows> allFollowRels = followsDao.findFollowsByUserToFollowId(userToUnfollowId);
 
-      return userToUnfollowId;
+      for (Follows rel : allFollowRels) {
+         if (rel.getFollower().getId() == userId) {
+            followsDao.delete(rel);
+            return user.getUsername() + " unfollowed " + userToUnfollow.getUsername();
```
DPigeon, line 303: Same as above.

DPigeon (changes requested):
Amazing job Hossein! I reviewed the backend part of this PR. Left some comments.
We'll need some unit and integration tests. Let me know if you need help with them.

backend/src/main/java/com/capstone/moneytree/model/SanitizedUser.java
EspressoCode: Very good idea, I agree, that is the right way of doing that but didn't come to my mind, I will change it in my next commit.

EspressoCode: Very good idea, I agree, that is the right way of doing that but didn't come to my mind, I will change it in my next commit.

EspressoCode: Very good idea, I agree, that is the right way of doing that but didn't come to my mind, I will change it in my next commit.

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
EspressoCode: Honestly I did not know if it's not good practice, I will revert it back to what it was before in my next commit.

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
EspressoCode: same as above there lol

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
EspressoCode: Actually for the frontend part we only care about response code to be 2XX so I think it would be better to revert it to Long id.

EspressoCode:
@David 
Please let me refactor the code tomorrow, I will make requested changes and will then resolve conflicts. For tests I will definitely need help, I thought it would be easy to fix them but I encountered problems that I think it would be better to pair program to fix them, let me know when you are free to do so, thanks.

DPigeon:
> @David 
> Please let me refactor the code tomorrow, I will make requested changes and will then resolve conflicts. For tests I will definitely need help, I thought it would be easy to fix them but I encountered problems that I think it would be better to pair program to fix them, let me know when you are free to do so, thanks.

Sure! Let me know when. We can pair program.

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
DPigeon: Before, it was only giving the user ID that it followed. Maybe giving the name instead would be useful or even the whole user object. Not sure what the frontend needs.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
DPigeon: Yes. Read the comment up there about that.

backend/src/main/java/com/capstone/moneytree/model/node/User.java
```diff
@@ -113,4 +94,13 @@ public void owns(Stock stock) {
         }
         this.getStocks().add(stock);
     }
-}    
\ No newline at end of file
+
+    public static User sanitizeUser(User user) {
+        User sanitizedUser = User.builder().firstName(user.getFirstName()).lastName(user.getLastName())
```
DPigeon, line 99: Minor comment here: could you put an attribute per line? So that it is better to read for the future. For instance:
```
User sanitizedUser = User.builder().
                                         .firstname(...)
                                         .lastName(...)
                                         ...
                                        .build();
```

backend/src/test/groovy/com/capstone/moneytree/model/SanitizedUserTest.groovy
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Unit Tests for the SanitizedUser model.
+ */
+class SanitizedUserTest extends Specification {
+    private static SanitizedUser sanitizedUser;
+
+    def setup() {
+        User user = createUser(1, "test@test.com", "username", "password", "Jake", "Moreau", "123B")
+        sanitizedUser = new SanitizedUser(user)
```
DPigeon, line 14: Did you do a test inside this Groovy class? I don't see one. Maybe my browser is being weird.

DPigeon (changes requested):
Looks good Hossein! I left some other comments. Good job on part 2 of this feature! 100% coverage is pretty good on the backend.

backend/src/test/groovy/com/capstone/moneytree/model/SanitizedUserTest.groovy
EspressoCode: My bad, I forgot that, just added it in my new commit.

DPigeon (approved):
Approved. Nice changes!

Issue #309, opened 2021-02-22
Add user's portfolio value over time endpoint


Alessandro100:
As a developer I would like to have an endpoint to get the user's portfolio value over time

NOTES:
- https://alpaca.markets/docs/api-documentation/api-v2/portfolio-history/

**Acceptance Criteria**
- Add the portfolio history either directly to the user when calling GET user or as a separate endpoint (your design decision)

PR #316, opened 2021-03-02
#309: Add User Portfolio Over Time

DPigeon:
# Related Issue
- #309 

# Proposed changes
- Add the alpaca GET portfolio at `/api/alpaca/portfolio/userId={userId}&period={periodLength}&unit={periodUnit}&timeframe={timeFrame}&dateend={dateEnd}&extended={extendedHours}`
- More information on the query parameters [here](https://alpaca.markets/docs/api-documentation/api-v2/portfolio-history/). Also see the additional information
- Added Alpaca IT and Unit tests

# Additional Info
- User ID is a string
- Period length is an integer
- Period Unit is either D (day), W (week), month (M) or year (A)
- Time Frame is either 1Min, 5Min, 15Min, 1H or 1D
- DateEnd is a date in 'YYYY-MM-DD' format
- Extended hours is either 'true' or 'false'

# Checklist (if applicable)
- [x] Tests

# Reviewer(s)
 - @razine-bensari 
 - @EspressoCode 
 - @marwan-ayadi (for frontend)

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/controller/AlpacaController.java
+++ b/backend/src/main/java/com/capstone/moneytree/controller/AlpacaController.java
@@ -40,8 +40,8 @@ public class AlpacaController {
     *
     * @return ResponseEntity of Account.
     */
-   @GetMapping("/account")
-   public ResponseEntity<Account> getAccount(String userId) {
+   @GetMapping("/account/{userId}")
+   public ResponseEntity<Account> getAccount(@PathVariable(name = "userId") @Valid @NotBlank String userId) {
       Account account = marketInteractionsFacade.getAccount(userId);
 
       return ResponseEntity.ok(account);
@@ -52,8 +52,8 @@ public class AlpacaController {
     *
     * @return ResponseEntity of Position List.
     */
-   @GetMapping("/positions")
-   public ResponseEntity<List<Position>> getPositions(String userId) {
+   @GetMapping("/positions/{userId}")
+   public ResponseEntity<List<Position>> getPositions(@PathVariable(name = "userId") @Valid @NotBlank String userId) {
       List<Position> positions = marketInteractionsFacade.getOpenPositions(userId);
 
       return ResponseEntity.ok(positions);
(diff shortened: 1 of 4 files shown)


EspressoCode:
Good job on implementation and the tests, I left you a small comment.

backend/src/main/java/com/capstone/moneytree/controller/AlpacaController.java
```diff
@@ -89,6 +89,17 @@ public AlpacaController(MarketInteractionsFacade marketInteractionsFacade, SimpM
...
    }
 
+   /**
+    *  Gets the market status (open/closed).
+    *
+    * @return market status.
+    */
+   @GetMapping("/market-status/{userId}")
+   public ResponseEntity<Clock> getMarketClock(@PathVariable(name = "userId") @Valid @NotBlank String userId) {
+      Clock marketClock = marketInteractionsFacade.getMarketClock(userId);
```
EspressoCode, line 99: Before implementation of the websocket, we did not have the userId as a parameter for getClock method, but now I see why Alessandro added it, he needed to create the alpaca session using that Id. The market-status is not dependent on a user, so I think we can keep the getclock method with its parameter, but maybe we can remove the {userId} from the path as it can be confusing. Let me know about your decision and I can reflect the changes in the frontend if you like.

DPigeon: I don't know much about that endpoint. I had in mind that AlpacaAPI would complain if no oAuthToken was attached to it. I'd have to try without the oAuthToken and see if it still does the job. All the other endpoint requires a user ID though.

EspressoCode: I created that endpoint, and it was working with only the Alpaca API key and the Alpaca Secret Key (tried it both on postman and in our app). More information is available here:
https://alpaca.markets/docs/api-documentation/api-v2/clock/

EspressoCode: Sorry David, you were right, I see that it is not working anymore without the Alpaca session, so please keep what you have, and mark this as resolved, thanks.

DPigeon: I will. No worries!

EspressoCode (approved):

Issue #274, opened 2021-01-26
Implement follower search bar functionality


Alessandro100:
As a user I would like to be able to quickly and easily follow / unfollow users

Notes:
- [Discussion] how should this work from the backend perspective? Should we return all the users (only name, pic, score)?

Changes from Figma:
- So as we have it right now, the search bar is a separate popup, but I think it would be more convenient to the user if it was directly on the UI like we did it for the stock search

**Acceptance Criteria**
- Ability to search a user through the search bar
- When searching a user, ability to see their: username, name, score, pic and their follow status (following / follow)
- Ability to follow / unfollow users from this search

![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/485f5d98-6e5d-472f-8969-ee9397fe4c96)

PR #318, opened 2021-03-05
Add users in search bar + tests

marwan-ayadi:
# Related Issue
#274

# Proposed changes
- Add users to search bar
- Add unit tests 

# Reviewer(s)
 @EspressoCode

### DIFF
--- a/client/src/app/components/header/header.component.ts
+++ b/client/src/app/components/header/header.component.ts
@@ -1,4 +1,6 @@
-import { Component, Input } from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
+import { UserSearch } from 'src/app/interfaces/userSearch';
+
 import { StoreFacadeService } from '../../store/store-facade.service';
 
 @Component({
@@ -6,10 +8,23 @@ import { StoreFacadeService } from '../../store/store-facade.service';
   templateUrl: './header.component.html',
   styleUrls: ['./header.component.scss'],
 })
-export class HeaderComponent {
+export class HeaderComponent implements OnInit {
   @Input() userPhotoURL: string;
+  userSearch$ = this.storeFacade.userSearch$;
+  showSearchBar = false;
   constructor(private storeFacade: StoreFacadeService) {}
+  ngOnInit(): void {
+    this.storeFacade.loadUserSearchList();
 
+    this.userSearch$.subscribe((userList: UserSearch[]) => {
+      // Loading search user list
+      if (!!userList) {
+        this.showSearchBar = true;
+      } else {
+        this.showSearchBar = false;
+      }
+    });
+  }
   logout(): void {
     this.storeFacade.logCurrentUserOut();
   }
--- a/client/src/app/components/stock-search/stock-search.component.html
+++ b/client/src/app/components/stock-search/stock-search.component.html
@@ -12,14 +12,51 @@
   #auto="matAutocomplete"
   (optionActivated)="getActiveOption($event)"
 >
-  <mat-option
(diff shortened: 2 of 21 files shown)


client/src/app/components/header/header.component.ts
```diff
@@ -1,15 +1,30 @@
...
 
+    this.userSearch$.subscribe((userList: UserSearch[]) => {
+      // Loading search user list
+      if (!!userList) {
+        this.showSearchBar = true;
+      } else {
+        this.showSearchBar = false;
+      }
+    });
+  }
```
EspressoCode, line 27: This can be changed to:
this.showSearchBar = !!userList ? true : false;

client/src/app/components/stock-search/stock-search.component.html
```diff
@@ -12,14 +12,51 @@
...
+      <span>Users</span>
+    </div>
+  </div>
+  <mat-divider></mat-divider>
+  <div class="custom-scroll">
+    <mat-option
+      *ngFor="let item of searchResults"
+      [value]="item"
+      [routerLink]="redirectSearch(item.type) + item.id"
+      (click)="query = item.name"
```
EspressoCode, line 44: Not sure if it worth having a function for this click, I like this one too!

client/src/app/interfaces/userSearch.ts
```diff
@@ -0,0 +1,8 @@
+export interface UserSearch {
```
EspressoCode, line 1: I think we could use our existing User interface? 
user interface

client/src/app/store/reducers/app.reducer.ts
```diff
@@ -14,6 +15,7 @@ export interface State {
   errorMessage: AppError;
   followers: User[];
   followings: User[];
+  searchUserList: UserSearch[];
```
EspressoCode, line 18: Do we need this to be added to our state? @Alessandro100 what do you think?

EspressoCode: and same thing about StockHistoricalData in the other PR, do we need to keep it in our state?

EspressoCode (changes requested):
Very good job on this story, good tests, nice implementation! I added some comments.

EspressoCode:
I want to add 2 more things, first one I thing should be fixed but the second is more of an improvement:

1) When searching for users, we need to remove the current user from the search results (comparing to the current user's email address because it's unique) and not return the current user itself.
ex: if I signed in as Hossein, I shouldn't see Hossein on the list.

2) We can add username to this feature maybe in future as an improvement.

client/src/app/components/header/header.component.ts
Alessandro100: it could actually just be ```this.showSearchBar = !!userList```

client/src/app/components/stock-search/stock-search.component.html
marwan-ayadi: Its needed because i have to identify if it is a stock or a user

client/src/app/interfaces/userSearch.ts
marwan-ayadi: I could but it is better for me to have another interface since in this interface the id is the username (I need the same format as the search stock results

client/src/app/components/header/header.component.ts
EspressoCode: > it could actually just be `this.showSearchBar = !!userList`

that's true 😄

EspressoCode (approved):

Issue #325, opened 2021-03-10
Edit profile page not working properly


EspressoCode:
**Describe the bug**
The update cover photo button on the edit profile page is not working as expected.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the edit profile page for any user.
2. Click on edit cover photo and try to change the cover photo.
3. Save changes and open the page again 
4. You will see that the picture you selected replaces the user photo instead of the cover photo

**Expected behaviour**
Cover photo and user profile photo should be updated correctly and without affecting the other.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version 88
 
**Additional context**
The reason is the replacement of old code by new code in this commit:
https://github.com/DPigeon/Money-Tree/pull/280/commits/62fca7c0c771602038c0d6b46904bf0848b45040
After discussing it with @abdulmansour, a part of the deleted code was added back but the missing part still affects this feature.

PR #326, opened 2021-03-10
#325 revert back the removed code - edit profile BE

EspressoCode:
# Related Issue
- #325 

# Proposed changes
- Reverts the working code in DefaultUserService which was replaced by new code and caused the edit profile not to work properly
- Improved edit profile, by adding a new default cover photo to S3 and used it for when a new user is being created
- Fixed the failing backend test

# Additional Info
- I uploaded the new default cover photo and named it "COVER"+ the default profile picture name so that there would be no need to add a new env variable for that.

# Reviewer(s)
 - @DPigeon

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
+++ b/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
@@ -115,6 +115,7 @@ public class DefaultUserService implements UserService {
         String encryptedPassword = encryptData(password);
         user.setPassword(encryptedPassword);
         user.setAvatarURL(String.format("https://%s.s3.amazonaws.com/%s", bucketName, DEFAULT_PROFILE_NAME));
+        user.setCoverPhotoURL(String.format("https://%s.s3.amazonaws.com/%s", bucketName, "COVER-"  + DEFAULT_PROFILE_NAME));
 
         userDao.save(user);
 
@@ -154,6 +155,10 @@ public class DefaultUserService implements UserService {
             userToUpdate.setAvatarURL(user.getAvatarURL());
         }
 
+        if (user.getCoverPhotoURL() != null) {
+            userToUpdate.setCoverPhotoURL(user.getCoverPhotoURL());
+        }
+
         if (user.getScore() != null) {
             userToUpdate.setScore(user.getScore());
         }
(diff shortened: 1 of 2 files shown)


backend/src/test/groovy/com/capstone/moneytree/service/DefaultUserServiceTest.groovy
```diff
@@ -111,7 +111,7 @@ class DefaultUserServiceTest extends Specification {
         userDao.save() >> user
 
         when:
-        User userResponse = userService.editUserProfilePicture(user, image, "test")
+        User userResponse = userService.editUserProfilePicture(user, image, "avatarURL")
```
DPigeon, line 114: Since cases were added into DefaultUserService in editUserProfilePicture method, add a test for "coverPhotoURL" and "default" selection. Those paths are not covered by the tests.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -154,6 +155,10 @@ public User updateUser(User userToUpdate, User user) {
             userToUpdate.setAvatarURL(user.getAvatarURL());
         }
 
+        if (user.getCoverPhotoURL() != null) {
+            userToUpdate.setCoverPhotoURL(user.getCoverPhotoURL());
```
DPigeon, line 159: This was added and needs a test for this path. Need to keep the coverage high :P

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -167,7 +172,11 @@ public User updateUser(User userToUpdate, User user) {
         }
 
         if (user.getPassword() != null) {
-            userToUpdate.setPassword(user.getPassword());
+            userToUpdate.setPassword(encryptData(user.getPassword()));
+        }
+
+        if (user.getBiography() != null) {
```
DPigeon, line 178: This was added and needs a test for this path. Need to keep the coverage high :P

DPigeon (changes requested):
I added some comments about testing. Good work!

backend/src/test/groovy/com/capstone/moneytree/service/DefaultUserServiceTest.groovy
EspressoCode: You are right, I am adding tests right now :-)

DPigeon (approved):

PR #327, opened 2021-03-10
Feature#279/view transaction and status

arthur-del:
https://app.zenhub.com/workspaces/moneytree-board-5f5bcf1c7ea48f001c28c973/issues/dpigeon/money-tree/279

# Proposed changes
Added profile page
Added transaction history component

# Additional Info
Transaction history unit test could use some love with ngOnChange test (stuff im unsure of)
Profile test indicates that transaction component has low test coverage (stuff im unsure of)
I had to comment out some lines of the backend to get the transactions to work, should I push that? @EspressoCode 

# Reviewer(s)
 @Alessandro100 
 @EspressoCode


DPigeon:
Hey @arthur-del it seems like you don't have the latest dev branch merged into your branch. You're getting conflicts and the CI isn't running because of that.

EspressoCode:
> Hey @arthur-del it seems like you don't have the latest dev branch merged into your branch. You're getting conflicts and the CI isn't running because of that.

Hi David, I asked Artur to comment out 3 lines in MarketInteractionFacade that were causing errors for transaction updates. The conflict is only in that file, I will take care of that merge conflict, and my pr for fixing the transaction updates (tomorrow) will replace that part again!

client/src/app/components/transaction-history/transaction-history.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  styleUrls: ['./transaction-history.component.scss']
+})
+export class TransactionHistoryComponent implements OnInit, OnChanges {
+  @Input() userTransactions: Transaction[];
+  transactionHistory: Transaction[];
+  constructor() { }
+
+  ngOnInit(): void {
+
+  }
```
Alessandro100, line 16: If this isn't being used, you can delete it

client/src/app/components/transaction-history/transaction-history.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  ngOnChanges(): void {
+    if (this.userTransactions) { this.transactionHistory = this.userTransactions; }
+    console.log('Transaction history: ', this.transactionHistory);
```
Alessandro100, line 19: remove console log

client/src/app/components/transaction-history/transaction-history.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+  transactionFormatStart(transaction: Transaction): string {
+    let printedTransaction = '';
+    if (transaction) {
+      transaction.type === 'MARKET_BUY' ? printedTransaction += 'Bought ' : printedTransaction += 'Sold ';
```
Alessandro100, line 24: this line can be re-written as below. It does the same thing, it's just more readable

```printedTransaction += transaction.type === 'MARKET_BUY' ?  'Bought ' : 'Sold ';```

File: client/src/app/components/transaction-history/transaction-history.component.ts [hunk 4]
```diff
@@ -0,0 +1,34 @@
...
+export class TransactionHistoryComponent implements OnInit, OnChanges {
+  @Input() userTransactions: Transaction[];
+  transactionHistory: Transaction[];
+  constructor() { }
+
+  ngOnInit(): void {
+
+  }
+  ngOnChanges(): void {
+    if (this.userTransactions) { this.transactionHistory = this.userTransactions; }
```
Alessandro100, line 18: Can you explain why you need an ngOnChanges?
Secondly, if the whole purpose of the ```transactionHistory``` variable is to be a copy of the ```userTransactions``` input, why don't you just use the ```userTransactionsInput```?

client/src/app/pages/home/home.component.ts
```diff
@@ -50,6 +53,12 @@ export class HomeComponent implements OnInit {
       }
     });
   }
+
+  goToProfile(): void {  // click -> navigate to profile/username
+    this.router.navigate(['/profile/' + this.currentUser.username])
+    .then(() => this.storeFacade.loadUserTransactions(this.currentUser.id));
```
Alessandro100, line 59: we shouldn't load the transactions when the user navigates to the profile. We should load them when we load the user. The transactions will be updated with the frontend webhook. Let me know if you need help on where to call this

client/src/app/pages/profile/profile.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  styleUrls: ['./profile.component.scss']
+})
+export class ProfileComponent implements OnInit {
+  userTransactions$ = this.storeFacade.userTransactions$;
+  constructor(
+    private storeFacade: StoreFacadeService
+  ) { }
+  ngOnInit(): void {
+    this.userTransactions$ = this.storeFacade.userTransactions$;
+    console.log('User Info in profile component', this.userTransactions$);
```
Alessandro100, line 16: remove console logs

Alessandro100 (commented):
Overall really nicely done, I can see your code improving since the beginning. I left a few comments regarding code quality. One thing that wasn't considered was the state of the Transaction history component if there are 0 transactions. I know it's not in figma, but we could at least add "You do not have any transactions" something as simple as that. Also, there seems to be a bug with transactions (or maybe it was on my end) but I couldn't see any transaction status besides PENDING so I couldn't test the functionality fully

Alessandro100:
Also, did you consider viewing other user's profiles?

client/src/app/pages/home/home.component.ts
EspressoCode: hey @Alessandro100, I wanted to discuss this with you, actually there's a problem with how we use state right now. We are losing state info whenever we refresh or navigate to a new page. Let's say we navigate from home page to stock detail page, we need to load state user, stock etc, and if we don't load followers for example, we'll have a null for followers in the state.
I think we should presist the whole state all the time, maybe by using NgRx local storage?

Alessandro100: State does get lost when we refresh the page. This is why if a user logs in, we have their user id stored in local storage. Then when the page refreshes, it will check if there is an id in local storage -> if there is, we will GET the user. Now that the user is separated from their transactions and followers, we will need to make the same calls for those over there. In terms of losing state when navigating, that shouldn't be the case.

Alessandro100: Whats going on here on line 59 is everytime we go to the profile page, we will GET the transactions. This can lead to many unused calls

Alessandro100 (approved):
Nice changes and code

Issue #279, opened 2021-01-26
View transactions and their status


Alessandro100:
As a user I would like to see my past transactions 

Note:
- the data for the transactions should be in the user object (the backend already returns it)

**Acceptance Criteria**
- The UI looks like the image under
- It accommodates different styles for different types of orders
- It displays real transactions order with the newest being first
- clicking on the ticker brings to the adequate stock page

![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/85416cfa-5427-4dbe-b758-0029033debc2)

Alessandro100:
@arthur-del  this story has a system test for it

PR #348, opened 2021-03-16
Enhancements#1/transaction history fixes and refactor

arthur-del:
#279  -> quality changes
# Proposed changes
Created a dataFormatter that handles user and transactions formatting
Created a default state when user has no transactions

# Reviewer(s)
 - @Alessandro100


Alessandro100 (changes requested):
The code looks good but can you fix the code smells, they should be quick to fix. Also the coverage is pretty bad, can you add the tests of the formatters to the new formatter file. Nice refactor!

Alessandro100 (approved):
nice changes

Issue #346, opened 2021-03-15
Update the user's score when they sell a stock


Alessandro100:
As a user I would like my score to update as soon as I sell a stock

Score Update Logic:
- Check % it went up or down and adjust like that
ex: If I bought AAPL at 100$ and sold at 110$ -> +10 points
ex: If I bought AAPL at 100$ and sold at 90$ -> -10 points

Notes:
- I know this logic is flawed and that there are better ways to calculate score, but for now this is the easiest

**Acceptance Criteria**
- [x] When the user SELLS a stock, update their score in the database
- [x] Assure that the default score for a user is 0

PR #350, opened 2021-03-20
#346: Update Score When User Sell Stock

DPigeon:
# Related Issue
- #346 

# Proposed changes
- User's score defaults to 0 when registered
- Updating and calculating the user's score when a transaction is processed
- Added 6 tests for all possible paths

# Additional Info
- A user will have its score updated if a transaction bought with the same symbol is found within the database. Then, the score will be nevative or positive depending on its performance in selling. If sold at $40 but bought at $10, +30 points. If sold at $20 but bought at $60, -40 points.

# Checklist (if applicable)
- [x] Tests

# Reviewer(s)
 - BackendTeam

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultTransactionService.java
+++ b/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultTransactionService.java
@@ -152,7 +152,11 @@ public class DefaultTransactionService implements TransactionService {
                break;
             }
          }
-         calculateScoreAndUpdate(boughtPrice, transaction.getTotal(), user);
+
+         float soldPrice = transaction.getTotal();
+         if (boughtPrice != soldPrice) { // No need to update score if prices are the same
+            calculateScoreAndUpdate(boughtPrice, soldPrice, user);
+         }
       }
    }
 
--- a/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultTransactionService.java
+++ b/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultTransactionService.java
@@ -7,7 +7,6 @@ import com.capstone.moneytree.dao.*;
 import com.capstone.moneytree.exception.AlpacaException;
 
 import com.capstone.moneytree.exception.EntityNotFoundException;
-import com.capstone.moneytree.model.relationship.Owns;
 import net.jacobpeterson.alpaca.enums.OrderSide;
 import net.jacobpeterson.alpaca.enums.OrderTimeInForce;
 
@@ -105,9 +104,8 @@ public class DefaultTransactionService implements TransactionService {
          LOGGER.info("Executed order {}", alpacaOrder.getClientOrderId());
 
          transaction = constructTransactionFromOrder(alpacaOrder);
-         transactionDao.save(transaction);
-
          updateUserScore(order, user, transaction);
+         transactionDao.save(transaction);
 
          /*
           * create and save two relationships: Made and ToFulfill. User Made Transaction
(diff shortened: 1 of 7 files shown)


backend/src/main/java/com/capstone/moneytree/service/impl/DefaultTransactionService.java
```diff
@@ -127,12 +127,46 @@ private void executeTransaction(String alpacaKey, Order order, User user) {
...
+   }
+
+   // ISSUE-346
+   public void updateUserScore(Order order, User user, Transaction transaction) {
+      if (order.getSide().equals("sell")) {
+         float boughtPrice = 0;
+         List<Made> userMade = madeDao.findByUserId(user.getId());
+         for (Made made : userMade) {
+            Transaction oldTransaction = made.getTransaction();
+            if (oldTransaction.getSymbol().equals(transaction.getSymbol())) {
```
EspressoCode, line 148: If the user bought AAPL stocks in different transactions (different days and different quantities and prices), then we will be calculating his score based on his last transaction only. I think maybe we can calculate the average purchase price based on quantity, and then calculate the price with that number. for example, if the user has 3 transactions with this info:
AAPL quantity: 5, price = 10 
AAPL quantity: 3, price = 20
AAPL quantity: 1, price = 5  
Average price = (10*5 + 20*3 +1*5) / (5+3+1)   = 12.78

EspressoCode (changes requested):
Nice changes David, I left you a small comment :-)

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultTransactionService.java
DPigeon: I know that Alessandro wanted something really simple because of time constraints for this project and features that we have to pull out. I also know that this only checks the last transaction. I had in mind before starting this feature to do something like you mentioned with stock quantity purchased. However, Alessandro told me not to worry about it. So I'm not sure what's the idea here anymore.

@Alessandro100 what do you think?

DPigeon: Edit: Actually I think you're right Hossein. Your solution is better. I'll work on it.

DPigeon: Just pushed some new changes. I followed that formula since I think it is better if the user has more than one transaction done.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultTransactionService.java
```diff
@@ -140,33 +140,37 @@ private Transaction constructTransactionFromOrder(Order alpacaOrder) {
...
+         float sumQty = 0;
          List<Made> userMade = madeDao.findByUserId(user.getId());
          for (Made made : userMade) {
-            Transaction oldTransaction = made.getTransaction();
-            if (oldTransaction.getSymbol().equals(transaction.getSymbol())) {
-               boughtPrice = oldTransaction.getTotal();
-               break;
+            Transaction otherTransaction = made.getTransaction();
+            if (otherTransaction.getSymbol().equals(transaction.getSymbol())) {
+               sumQtyPrice = sumQtyPrice + otherTransaction.getTotal() * otherTransaction.getQuantity();
```
EspressoCode, line 151: very nice, thanks for this change, this is exactly what I meant.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultTransactionService.java
EspressoCode: Very nice David, thanks for making the changes and for changing the tests as well. I approved this PR.

EspressoCode (approved):

Issue #268, opened 2021-01-26
Add additional social information to the stock endpoint


Alessandro100:
Notes
- "Top Investors who own this stock": Return the profile of user's who's scores are in the top 10%
- "Followers who also own this stock": Return the profile of user's who the current user follows AND own the stock
- "People who own this stock also own": Return the top 5 stocks that people who own this 
        - example: let's say 5 people own AAPL, and 4/5 of those people also own MSFT, then it would return [MSFT 80%]

Extra Notes
- These queries should be neo4j efficient, we can discuss the functionality of them
- This story might be done later in the sprint as we will get the information later (followers, updating scores, etc)

**Acceptance Criteria**
- Add "Top Investors who own this stock" to the GET stock endpoint
- Add "Followers who also own this stock" to the GET stock endpoint
- Add "People who own this stock also own" to the GET stock endpoint

DPigeon:
Part 1 and part 2 were merged. Is there a part 3 coming from @abdulmansour ?

PR #351, opened 2021-03-24
Adding top 10 users based on score and stock owned

razine-bensari:
# Related Issue
- #268 part 1

# Proposed changes
- add endpoint for frontend to  get top 10% users of a certain stock

# Additional Info
- in order for it to work, we need some good data in the DB

# Checklist (if applicable)
- Api test, 
-  integration test

# Video or links (if applicable)
 - url

# Reviewer(s)
 - @abdulmansour 
 - @Alessandro100 
 - @DPigeon

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/model/node/User.java
+++ b/backend/src/main/java/com/capstone/moneytree/model/node/User.java
@@ -1,6 +1,7 @@
 package com.capstone.moneytree.model.node;
 
 import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.jetbrains.annotations.NotNull;
 import org.neo4j.ogm.annotation.NodeEntity;
 
 import com.capstone.moneytree.model.Entity;
@@ -15,7 +16,7 @@ import lombok.NoArgsConstructor;
 @Builder
 @NoArgsConstructor
 @AllArgsConstructor
-public class User extends Entity {
+public class User extends Entity implements Comparable<User> {
 
     String firstName;
 
@@ -67,5 +68,10 @@ public class User extends Entity {
         }
 
         return false;
+    }
+
+    @Override
+    public int compareTo(@NotNull User user) {
+        return getScore().compareTo(user.getScore());
     }
 }
--- a/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
+++ b/backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
@@ -1,5 +1,8 @@
 package com.capstone.moneytree.service.impl;
 
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toList;
+
 import javax.security.auth.login.CredentialNotFoundException;
 
 import com.capstone.moneytree.ActiveProfile;
@@ -35,6 +38,7 @@ import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
@@ -64,6 +68,7 @@ public class DefaultUserService implements UserService {
     ActiveProfile activeProfile;
 
     private final UserDao userDao;
+    private final OwnsDao ownsDao;
     private final FollowsDao followsDao;
     private final ValidatorFactory validatorFactory;
     private final MoneyTreePasswordEncryption passwordEncryption;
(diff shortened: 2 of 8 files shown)


backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -374,6 +380,32 @@ public void deleteUserByEmail(String email) {
...
+        // reverse to get biggest score first
+        Collections.reverse(users);
+
+        int top10percent = users.size() / 10; // rounded down for conservative results
+
+        return users.stream()
+                .limit(top10percent)
+                .collect(toList());
+    }
+
```
razine-bensari, line 408: For now, we can afford to do the query on server side since the number of users/relationship is low and not a bottleneck. The moment we hit more users, we would need to transfer this query to database layer

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -384,6 +416,8 @@ public UserValidator getUserValidator() {
         return validatorFactory.getUserValidator();
     }
 
+
+
```
razine-bensari, line 420: sorry about that guys

backend/src/main/java/com/capstone/moneytree/model/node/User.java
```diff
@@ -68,4 +69,9 @@ public boolean equals(Object object) {
 
         return false;
     }
+
+    @Override
+    public int compareTo(@NotNull User user) {
+        return getScore().compareTo(user.getScore());
+    }
```
razine-bensari, line 76: needed for sorting purposes

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
razine-bensari: I also floor down the top 10% for conservative values so if we have 95 users, then the top 10% is 9 users

DPigeon: Yeah this will be server heavy. Let`s see how many users we get.

DPigeon (commented):
Nice changes! I`ll approve once the checks are passing.

DPigeon (approved):

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -374,6 +380,32 @@ public void deleteUserByEmail(String email) {
...
+        if (users.isEmpty()) {
+            return emptyList();
+        }
+
+        // sort user by ascending score order
+        Collections.sort(users);
+
+        // reverse to get biggest score first
+        Collections.reverse(users);
```
wltrf, line 400: Everything here is fine and works. I'd just like to mention that it could be reduced:
- Instead of the .forEach after the stream, you can use .map(Owns::getUser) after the .filter
- You can then sort and reverse inside the stream, since you are now dealing with the users, as follows:
```java
.sorted(Comparator.comparing(User::getScore).reversed())
```
And since you made the User class comparable on score, you can drop the getScore

wltrf (commented):
looks good, just need to fix the conflicts and you can address my comment if you feel like it.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -371,6 +376,32 @@ public void deleteUserByEmail(String email) {
     }
 
     @Override
+
+    public List<User> getTopUsers(String symbol) {
+        List<User> users = new ArrayList<>();
+        // get a list of all user who own a stock with this symbol
+        ownsDao.findAll().stream()
+                .filter(owns -> owns.getStock().getSymbol().equals(symbol))
```
EspressoCode, line 384: Hi Razine, I tried this branch and realized that there are 2 bugs:
1) I know that right now, when we create a user we dedicate a score=0 to him/her, but for when we are using old users, they all have null score, so the compare method will throw null exception for when we are comparing the scores. Please change this line to this:
`  .filter(owns -> owns.getStock().getSymbol().equals(symbol) && owns.getUser().getScore()!=null)`
It is hard to make transactions that are updated inside alpaca market hours, so this would save us from resetting the database users to avoid null pointer.

2) I tried the endpoint with AAPL stock, with a user that had two AAPL transactions, I mean he bought AAPL in 2 different transactions. The endpoint returned the user 2 times, so I think we need to either use set<User> to avoid repetitions in our response.

EspressoCode (changes requested):
Good job, I left you a comment.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
razine-bensari: Good catch, will add the changes

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
razine-bensari: Thks for the optimization, will add it

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
razine-bensari: Hey Hossein, for the second points, even if the user made the transaction 2 times, he still owns the stock no? since I sort using the Owns relationship. If he owns the stock two times it is not good no ?

EspressoCode: I meant, we should return the list of 10 distinct users, but I got 1 user two times in the response.

EspressoCode (approved):

PR #352, opened 2021-03-24
Issue/#268 followers who also own this stock

razine-bensari:
# Related Issue
- #268 part 2

#### Proposed changes
- change 1
- added an endpoint to get the list of followers who owns the same stock as the one provided in the id

#### Checklist (if applicable)
- integration test

#### Reviewer(s)
 - @DPigeon 
 - @Alessandro100 
 - @wltrf 
 - @EspressoCode

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/controller/UserController.java
+++ b/backend/src/main/java/com/capstone/moneytree/controller/UserController.java
@@ -197,4 +197,9 @@ public class UserController {
    public ResponseEntity<List<User>> getTopUserForStock(@PathVariable String symbol) {
       return ResponseEntity.ok(userService.getTopUsers(symbol));
    }
+
+   @GetMapping("{id}/owned_by_followers/{symbol}")
+   public ResponseEntity<List<User>> getStockOwnedByFollowers(@PathVariable Long id, @PathVariable String symbol) {
+      return ResponseEntity.ok(userService.getFollowersWhoOwnsTheStock(id, symbol));
+   }
 }
--- a/backend/src/main/java/com/capstone/moneytree/controller/UserController.java
+++ b/backend/src/main/java/com/capstone/moneytree/controller/UserController.java
@@ -201,9 +201,10 @@ public class UserController {
    @GetMapping("/{id}/owned_by_followers/{symbol}")
    public ResponseEntity<List<User>> getStockOwnedByFollowers(@PathVariable Long id, @PathVariable String symbol) {
       return ResponseEntity.ok(userService.getFollowersWhoOwnsTheStock(id, symbol));
+   }
 
    @GetMapping("/leaderboard")
-   public List<SanitizedUser> getLeaderboard(){
-      return userService.getLeaderboard();
+   public ResponseEntity<List<SanitizedUser>> getLeaderboard() {
+      return ResponseEntity.ok(userService.getLeaderboard());
    }
 }
--- a/backend/src/main/java/com/capstone/moneytree/controller/UserController.java
+++ b/backend/src/main/java/com/capstone/moneytree/controller/UserController.java
@@ -198,7 +198,7 @@ public class UserController {
       return ResponseEntity.ok(userService.getTopUsers(symbol));
    }
 
-   @GetMapping("{id}/owned_by_followers/{symbol}")
+   @GetMapping("/{id}/owned_by_followers/{symbol}")
    public ResponseEntity<List<User>> getStockOwnedByFollowers(@PathVariable Long id, @PathVariable String symbol) {
       return ResponseEntity.ok(userService.getFollowersWhoOwnsTheStock(id, symbol));
    }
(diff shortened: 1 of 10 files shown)


backend/src/main/java/com/capstone/moneytree/controller/UserController.java
```diff
@@ -197,4 +197,9 @@ public User login(@RequestBody User credentials) throws CredentialNotFoundExcept
    public ResponseEntity<List<User>> getTopUserForStock(@PathVariable String symbol) {
       return ResponseEntity.ok(userService.getTopUsers(symbol));
    }
+
+   @GetMapping("{id}/owned_by_followers/{symbol}")
+   public ResponseEntity<List<User>> getStockOwnedByFollowers(@PathVariable Long id, @PathVariable String symbol) {
+      return ResponseEntity.ok(userService.getFollowersWhoOwnsTheStock(id, symbol));
+   }
```
razine-bensari, line 204: We could settle on the naming of that endpoint not satisfied with what I chose

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -406,6 +411,18 @@ public void deleteUserByEmail(String email) {
...
+        List<User> followersWhoOwnsThisStock = new ArrayList<>();
+
+        User user = getUserById(id);
+        Stock stock = stockService.getStockBySymbol(symbol);
+        List<Follows> followers = followsDao.findByUserToFollowId(user.getId());
+
+        followers.forEach(follower -> addFollowerToList(follower, followersWhoOwnsThisStock, stock));
+        return followersWhoOwnsThisStock;
+    }
+
```
razine-bensari, line 425: This one can be kept in server side since followers are not too big. We could eventually limit it too. But if the list of followers is in the thousands, it becomes slow.

backend/src/test/groovy/com/capstone/moneytree/service/DefaultStockServiceTest.groovy
```diff
@@ -51,7 +60,34 @@ class DefaultStockServiceTest extends Specification {
         List<SanitizedStock> owns = stockService.getUserStocks(userId)
 
         then: "should get all user stocks"
-        assert owns != null
-        assert owns.size() == 1
+        owns != null
+        owns.size() == 1
+    }
+
+    @Ignore // for some reason this test fail. I have wasted too much time. if anyone knows how to fix it please do!!!
```
razine-bensari, line 67: please please please, I am baffled, I dont know why this one does not want to work!!!! 😭 😨

razine-bensari:
pointing it to feature branch for now since they are dependent on the change. But once the dependent branch is merged, the base branch of the PR should default to dev automatically

backend/src/test/groovy/com/capstone/moneytree/service/DefaultStockServiceTest.groovy
wltrf: It has a very weird behaviour. It seems the assertion in the "then" block is affecting the call in the "when" block, not sure why. 
Below is my rewrite of the test, which passes. I made the mocking less generic and more specific to the AAPL symbol since no other symbol should return the AAPL stock. I also replaced the called once assertion to a simple equality check, which seems more fitting.
```java
        given: "a stock"
        Stock stock0 = new Stock("AAPL", "Apple")

        and:
        stockDao.findBySymbol("AAPL") >> stock0

        when:
        Stock stock1 = stockService.getStockBySymbol("AAPL")

        then:
        assert stock1 == stock0
```

backend/src/main/java/com/capstone/moneytree/controller/UserController.java
wltrf: naming is fine, it clearly conveys what this endpoints returns even if it is not sexy.

wltrf (commented):
Looks good, just need to fix the failing test.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultStockService.java
```diff
@@ -20,6 +21,8 @@
 @Transactional
 public class DefaultStockService implements StockService {
 
+    private static final String STOCK_NOT_FOUND = "The requested stock was not found";
```
DPigeon, line 24: I think this line does not belong here. Add this into the [ExceptionMessage](https://github.com/DPigeon/Money-Tree/blob/dev/backend/src/main/java/com/capstone/moneytree/handler/ExceptionMessage.java#L4) class.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -60,12 +62,15 @@
     private static final String USER_ALREADY_FOLLOWED = "User already followed";
```
DPigeon, line 62: Yeah same thing here but this is from another story. It probably slipped through my eyes. :P

backend/src/test/groovy/com/capstone/moneytree/service/DefaultStockServiceTest.groovy
DPigeon: The above works. Thanks Walter!

DPigeon (changes requested):
Nice changes! Just a minor thing to do and we can push this to dev!

razine-bensari:
Thks @DPigeon  and @wltrf  will add the test fix and the pr comments

DPigeon (approved):

Issue #312, opened 2021-02-22
Buy/Sell Limit order not working


Alessandro100:
**Describe the bug**
When a user submits a limit buy order, it executes as a market order

**To Reproduce**
1. Go to stock detail page
2. Limit buy 1 stock
3. See console / alpaca saying it was executed as a market order

**Expected behavior**
It should be executed as a limit order

**Screenshots**
This are the params going to the  server 
![image](https://user-images.githubusercontent.com/18631060/108754277-ba93aa00-7513-11eb-9d7b-7f1220dbf23b.png)

This is whats coming back
![image](https://user-images.githubusercontent.com/18631060/108754320-c7180280-7513-11eb-938d-de43e2252b4f.png)

NOTE:
- same applies with sell

Bug reported by @Alessandro100

Alessandro100:
I just noticed that in the frontend we don't even pass the limit price. @DPigeon let me know the format of the limitprice variable for transaction

DPigeon:
> I just noticed that in the frontend we don't even pass the limit price. @DPigeon let me know the format of the limitprice variable for transaction

I haven't work on the buy/sell endpoints. I'dhave to understand what is happening here before answering. Maybe @razine-bensari or @abdulmansour can help?

razine-bensari:
for limit_* orders, the limit price needs to be provided in the payload, the boolean for market hours. Please refer to the order object for field names

PR #353, opened 2021-03-24
Fixing bug buy/sell limit orders

razine-bensari:
# Related Issue
- #312 

#### Proposed changes
- The issue was that we only placed market order and hardcoded the type. Now we support limit_buy, limit_sell, market_buy and market_sell

#### Additional Info
- for limit_* orders, the limit price needs to be provided in the paylod, the boolean for market hours. Please refer to the order object for field names

#### Checklist (if applicable)
- Did not break any unit tests. Please test locally if it works fine on your end. We stumbled upon a few issues where it was working well on some peoples machine but not all

#### Reviewer(s)
 - @DPigeon 
 - @Alessandro100 
 - @EspressoCode 
 - @abdulmansour 
 - @marwan-ayadi


EspressoCode:
Right now we update the balance using the alpaca balance but they give us a negative balance, for example when you create a user on Alpaca, the cash balance is negative: "cash": "-1024.6538"

Our frontend has a $100,000 dollar balance for when the user has no purchases (which was hardcoded and not accurate!)
then we do the first purchase and it changes to (that negative value) - (price you bought), and then because you have a negative value the buy button becomes disabled.

I think we can not rely on Alpaca, especially that I saw that they update the balance sometimes on a daily basis! My solution is to set the balance equal to 100,000 for when we create the user, and then in each transaction, we manually deduct/add to the balance. What do you think? @Alessandro100 @razine-bensari @DPigeon @wltrf 
I can make those changes, but I mentioned it here so that we don't merge this pr until I add those changes here, because this bug is related to this issue.

DPigeon:
> Right now we update the balance using the alpaca balance but they give us a negative balance, for example when you create a user on Alpaca, the cash balance is negative: "cash": "-1024.6538"
> 
> Our frontend has a $100,000 dollar balance for when the user has no purchases (which was hardcoded and not accurate!)
> then we do the first purchase and it changes to (that negative value) - (price you bought), and then because you have a negative value the buy button becomes disabled.
> 
> I think we can not rely on Alpaca, especially that I saw that they update the balance sometimes on a daily basis! My solution is to set the balance equal to 100,000 for when we create the user, and then in each transaction, we manually deduct/add to the balance. What do you think? @Alessandro100 @razine-bensari @DPigeon @wltrf
> I can make those changes, but I mentioned it here so that we don't merge this pr until I add those changes here, because this bug is related to this issue.

Good point. We need the user to have a 'base' balance. Totally agree.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultTransactionService.java
```diff
@@ -126,6 +131,20 @@ private void executeTransaction(String alpacaKey, Order order, User user) {
       }
    }
 
+   private Order executeRequest(AlpacaAPI api, Order order) throws AlpacaAPIRequestException {
+      if (MoneyTreeOrderType.MARKET_BUY.name().equals(obtainMoneyTreeOrderType(order).name()) ||
```
DPigeon, line 135: We'll have to test those two condition paths when the executeTransaction test is fixed.

razine-bensari: Added extra tests cases

razine-bensari:
> Right now we update the balance using the alpaca balance but they give us a negative balance, for example when you create a user on Alpaca, the cash balance is negative: "cash": "-1024.6538"
> 
> Our frontend has a $100,000 dollar balance for when the user has no purchases (which was hardcoded and not accurate!)
> then we do the first purchase and it changes to (that negative value) - (price you bought), and then because you have a negative value the buy button becomes disabled.
> 
> I think we can not rely on Alpaca, especially that I saw that they update the balance sometimes on a daily basis! My solution is to set the balance equal to 100,000 for when we create the user, and then in each transaction, we manually deduct/add to the balance. What do you think? @Alessandro100 @razine-bensari @DPigeon @wltrf
> I can make those changes, but I mentioned it here so that we don't merge this pr until I add those changes here, because this bug is related to this issue.

I agree, I actually dont like alpaca anymore, because of simple stuff like these. Its the little things man, the little things

EspressoCode:
> > Right now we update the balance using the alpaca balance but they give us a negative balance, for example when you create a user on Alpaca, the cash balance is negative: "cash": "-1024.6538"
> > Our frontend has a $100,000 dollar balance for when the user has no purchases (which was hardcoded and not accurate!)
> > then we do the first purchase and it changes to (that negative value) - (price you bought), and then because you have a negative value the buy button becomes disabled.
> > I think we can not rely on Alpaca, especially that I saw that they update the balance sometimes on a daily basis! My solution is to set the balance equal to 100,000 for when we create the user, and then in each transaction, we manually deduct/add to the balance. What do you think? @Alessandro100 @razine-bensari @DPigeon @wltrf
> > I can make those changes, but I mentioned it here so that we don't merge this pr until I add those changes here, because this bug is related to this issue.
> 
> I agree, I actually dont like alpaca anymore, because of simple stuff like these. Its the little things man, the little things

After talking to Alessandro, I realized that the negative balance is because we are using money-tree users over the same Alpaca account in our local, so for prod users that wouldn't be an issue. I think we are good to merge this PR :-)

DPigeon (approved):
Approved. And yes Alpaca is a struggle now that we're experimenting it more.

Issue #360, opened 2021-03-29
Refactoring Backend Components Part 1


DPigeon:
As a developer, I want to refactor some components in the backend so that our code looks cleaner.

There are many methods that can be refactored to simple ones. This story consists of working on those. Some test coverage will also be added there to fill the needs.

**Acceptance of Criteria:**
- [x] Refactor multiple old classes into simple one
- [x] Spot some duplicates or arguments in methods passed twice
- [x] Increase code coverage
- [x] Fix any other SonarCloud issue (https://sonarcloud.io/project/issues?id=money-tree_BACKEND&branch=dev&resolved=false&types=BUG)

DPigeon:
Question: while refactoring, I came across the decorator models (Orders). Are we ever going to use those?

PR #363, opened 2021-04-02
#360: Refactoring Backend Components [Part 1]

DPigeon:
# Related Issue
- #360 

# Proposed changes
- Added some new tests
- Fixed some SonarCloud bugs
- Refactored multiple methods (S3 Selection picture, Execute Transaction arguments, etc)
- Cleaned some old code
- A part 2 will be needed to finalize missing tests and possible new upcoming SonarCloud alerts.

# Additional Info
- Note: This PR has lots of changes and it is normal. I went over all files and did a "part 1" of what I thought could be improved.

# Checklist (if applicable)
- [x] Tests

# Reviewer(s)
 - BackendTeam


backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
```diff
@@ -57,24 +58,23 @@
     private static final String USER_CANT_BE_FOLLOWED_BY_ITSELF = "User cannot followed by itself";
 
     @Value("${alpaca.key.id}")
-    private String clientId;
```
wltrf, line 60: I think there used to be a code smell regarding this and private was added to fix it, but seems like it doesn't show up anymore so that's good.

backend/src/main/resources/application-dev.properties
```diff
@@ -36,4 +36,4 @@ spring.mail.port=587
 spring.mail.username=${SPRING_BOOT_EMAIL}
 spring.mail.password=${SPRING_BOOT_EMAIL_PASSWORD}
 spring.mail.properties.mail.smtp.auth=true
-spring.mail.properties.mail.smtp.starttls.enable=true
+spring.mail.properties.mail.smtp.starttls.enable=true
```
wltrf, line 39: We should keep the newline at the end of the file
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/729795

wltrf (approved):
Looks good, minor comment that can be addressed.

backend/src/main/java/com/capstone/moneytree/service/impl/DefaultUserService.java
DPigeon: Yeah I've seen some SonarCloud alerts coming back from time to time. If it comes back in future PRs I'll fix it.

backend/src/main/resources/application-dev.properties
DPigeon: Good point! I'll make the changes.

Issue #344, opened 2021-03-15
Implement Advanced Profile Information 


Alessandro100:
As a user I would like to see advanced profile information 

NOTE:
- The endpoint 'Get user by username' should contain all the information regarding user profile. Add all this information in that call and not in separate calls

**Acceptance Criteria**
- Show a piechart of the sectors the user is invested in
- Show the leaderboard rank

![image.png](https://images.zenhubusercontent.com/5e160ea2cbe5be5b659938e7/9d50c258-a681-4f25-bc1e-2a68d17bf43b)

PR #366, opened 2021-04-03
Feature/#344 advanced profile info

marwan-ayadi:
# Related Issue
- #344 

# Proposed changes
- Add pie chart for user's sectors of interest

# Additional Info
- need to buy new stocks (create new transaction to see the chart), old transaction will not display it because the industry was not set.

# Reviewer(s)
 - @EspressoCode

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/controller/UserController.java
+++ b/backend/src/main/java/com/capstone/moneytree/controller/UserController.java
@@ -99,6 +99,7 @@ public class UserController extends AbstractController {
       completeUserProfile.setFollowing(userService.getFollowings(user.getId()));
       completeUserProfile.setTransactions(transactionService.getUserTransactions(user.getId()));
       completeUserProfile.setOwnedStocks(stockService.getUserStocks(user.getId()));
+      completeUserProfile.setPercentile(userService.getUserPercentile(user.getUsername()));
       return completeUserProfile;
    }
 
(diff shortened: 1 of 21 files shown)


EspressoCode:
I tried it on my end and it works fine, so I approve it. Also, I fixed the failing tests on the backend and frontend.
image

We are aware of the problem of IEX cloud giving us random strings for industry and some other info, which is because of using the free version. So that can be fixed if we use the paid version and would not be an issue in that case.

@DPigeon  It is very important that we clear our db again for this pie chart to work, as we are now having two new properties for stocks and transactions, and our old stocks in db would not be replaced by new objects when we make a new purchase. For example if the user buy AAPL and we already have AAPL in db, the pie chart will not show up because the old AAPL has null value for industry.

EspressoCode (approved):

DPigeon:
> I tried it on my end and it works fine, so I approve it. Also, I fixed the failing tests on the backend and frontend.
> image
> 
> We are aware of the problem of IEX cloud giving us random strings for industry and some other info, which is because of using the free version. So that can be fixed if we use the paid version and would not be an issue in that case.
> 
> @DPigeon  It is very important that we clear our db again for this pie chart to work, as we are now having two new properties for stocks and transactions, and our old stocks in db would not be replaced by new objects when we make a new purchase. For example if the user buy AAPL and we already have AAPL in db, the pie chart will not show up because the old AAPL has null value for industry.
> 
> 

This will be fine on prod because we are always executing IT tests on the dev env.

Issue #368, opened 2021-04-03
IT Tests Cleanup Makes Dev Database Empty


DPigeon:
**Describe the bug**
When the integration tests are running locally or on the CI, some tests "deleteAll" entity persisted in the entire database.

**To Reproduce**
1. Create a user on dev.money-tree.tech
2. Run all backend tests
3. Look at the dev database
4. The user you created is also deleted because of the cleanup

**Expected behavior**
It should only delete needed users used in tests and not a user created on the frontend

**Desktop (please complete the following information):**
 - CI Pipeline

**Smartphone (please complete the following information):**
 - Linux

**Additional context**
This needs to be fixed ASAP to test on the DEV environment properly while others create PRs (making the CI pipeline run and deleting all entities on the dev db).

Bug reported by @DPigeon

PR #369, opened 2021-04-03
#368: Fix IT Tests Cleanup for Dev Database

DPigeon:
# Related Issue
- #368

# Proposed changes
- Properly clean up the database when multiple users are added to it
- Changed the order the entities are deleted when cleaning up (deleting from most recent created to last to properly delete the relationships)
- No more bug where we create a user to test on dev and it gets deleted every time the IT tests are running

# Additional Info
- After the CI passes, I'll report if my user entity was indeed kept like it should normally.

# Checklist (if applicable)
- [x] Tests (Tested on DEV and works, my user is not deleted anymore)

# Reviewer(s)
- BackendTeam

### DIFF
--- a/backend/src/test/groovy/com/capstone/moneytree/controller/UserControllerIT.groovy
+++ b/backend/src/test/groovy/com/capstone/moneytree/controller/UserControllerIT.groovy
@@ -137,8 +137,7 @@ class UserControllerIT extends Specification {
         }
 
         cleanup: "delete the created user"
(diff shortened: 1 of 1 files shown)


DPigeon:
Note: the tests all executed (IT tests) and I can still see my user entity on the Dev Database. ✔️

EspressoCode (approved):
This is exactly what we needed, removing relationships first and then removing the entities, good job David.

PR #375, opened 2021-04-03
making redux devtools available on dev

EspressoCode:
# Related Issue
- This would be a temporary solution to see state and appErrors while we are in at (the redux tool)

# Proposed changes
- it also removes the extra "/" for request URLs in uat

@DPigeon


DPigeon:
This is to test on the dev environment. It has to be put back to normal after.
Approved.

DPigeon (approved):

PR #381, opened 2021-04-05
#378 added api types to alpaca.properties

EspressoCode:
# Related Issue
- bug fix for #378

# Proposed changes
- added alpaca API types to properties

# Additional Info
- Based on the backend log on error on Digital Ocean:

Capture1

We have a null pointer error while we want to create the Alpaca session, but this is not about the null key, because as you can see, we are passing line 148 which means our key, secret or oAuth token is fine.

errrr

Line 149 is the problematic part which is complaining about API type, I added the 2 required API types in our properties files (based on https://github.com/Petersoj/alpaca-java#Configuration ). I hope by merging this maybe we get rid of that error.

@DPigeon And I don't understand why we don't have this issue on our local.

@DPigeon

### DIFF
--- a/backend/src/main/java/com/capstone/moneytree/facade/AlpacaSession.java
+++ b/backend/src/main/java/com/capstone/moneytree/facade/AlpacaSession.java
@@ -2,6 +2,9 @@ package com.capstone.moneytree.facade;
 
 
 import com.capstone.moneytree.handler.ExceptionMessage;
+import net.jacobpeterson.alpaca.enums.api.DataAPIType;
+import net.jacobpeterson.alpaca.enums.api.EndpointAPIType;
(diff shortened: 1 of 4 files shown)


EspressoCode:
and btw, this causes no errors on local, at least for when the market is closed 🤦🏻‍♂️

DPigeon:
Hey Hossein I pushed a fix to complete this PR. I think this is the reason why we get a nullpointer. Good point that you brought: ENDPOINTDATATYPE and DATATYPE were null. So I added them in the AlpacaAPI constructor.

Alessandro100 (approved):
looks good, hopefully this will fix the issue

DPigeon (approved):

PR #391, opened 2021-04-11
exclude service files from SonarCloud coverage

EspressoCode:
Service files are only responsible for calls to the backend and our Backend tests already took care of testing all the controllers, so there is no point to test them in the frontend again

 # Proposed changes
- Excluded service files from coverage in frontend
- Excluded  conf.js files
- SonarCloud settings on the SonarCloud website are now the same as our properties files

@David @Alessandro100


Alessandro100 (approved):
good initiative on removing test files that have no functionality

PR #392, opened 2021-04-11
final resolution of code smells in frontend

EspressoCode:
# Proposed changes
- final resolution of code smells in frontend

@Alessandro100


Alessandro100 (approved):
nice small refactor!

